PR #5, opened 2019-02-12
Make MicronautRequestHandler friendly to Spock mocks (fixes #4)

### LINKED ISSUE #4: Self-injecting MicronautRequestHandler prevents Spock mocks
The `MicronautRequestHandler` currently creates its `ApplicationContext` during the `handleRequest` method, and uses it to inject its dependencies immediately before calling the abstract `execute` method implemented by the subclass.

For anyone using Spock as their test framework, this breaks the ability to implement tests using mocked beans that are injected into the handler class.  I.e., in the current flow, a test is unable to:

1. Create and start an `ApplicationContext` that will contain the mock versions of beans
1. Extract the mocks from the started context, and call `MockUtil.attachMock` to attach them to the current specification
1. Declare mock/stub interactions on the mocks
1. Call the handler (which will then self-inject the mocks) prior to calling `execute`

I will submit a PR shortly with a proposed fix.

mmindenhall:
Proposed fix for #4, refactoring `MicronautRequestHandler` to allow testing via inject Spock mocks.  Includes a test to demonstrate and verify that the pattern now works.

### DIFF
--- a/function-aws/src/main/java/io/micronaut/function/aws/MicronautRequestHandler.java
+++ b/function-aws/src/main/java/io/micronaut/function/aws/MicronautRequestHandler.java
@@ -31,17 +31,19 @@ public abstract class MicronautRequestHandler<I, O> extends AbstractFunctionExec
 
     @Override
     public final O handleRequest(I input, Context context) {
-        ApplicationContext applicationContext = buildApplicationContext(context);
-        startEnvironment(applicationContext);
+        if (applicationContext == null) {
+            applicationContext = buildApplicationContext(context);
+        }
+        if (context != null) {
+            registerContextBeans(context, applicationContext);
+        }
         return applicationContext.inject(this).execute(input);
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


graemerocher:
Instead of this, you could just use Micronaut Test and `@MockBean`. Would this not solve the issue?

mmindenhall:
@graemerocher, I tried this, it doesn't work, and I don't think it _can_ work.  

Perhaps I'm missing something, but I think we have two different `ApplicationContext` instances being created here:  

1.  created by `@MicronautTest`, which understands `@MockBean` so it injects my mocks.
1. created by `MicronautRequestHandler.handleRequest`, to which it seems like `@MockBean` would be meaningless since that code path has no dependency on micronaut-test.

For this to work, `@MicronautTest` would somehow need to provide the `ApplicationContext` used by `MicronautRequestHandler`.

mmindenhall:
Any thoughts on this one @graemerocher?

graemerocher:
Will look into it soon.. thx

Issue #38, opened 2019-05-16
Server side rendering not working when using function-aws-api-proxy


jschwartz73:
Given the below controller action:
```
@Get("/render-html")
@Produces(MediaType.TEXT_HTML)
public ModelAndView getHtmlFromRenderEngine(Context context) {
        ModelAndView mav = new ModelAndView("home", CollectionUtils.mapOf("firstname", "Luke", "lastname", "Skywalker"));
        return mav;
}
```

When executing the action through AWS API Gateway (via the micronaut function-aws-api-proxy) the result will be:
```
{"view":"home","model":{"firstname":Luke,"lastname":"Skywalker"}}
```

The correct result should be rendered "home" template using the configured render engine.

graemerocher:
Thanks for the PR

PR #39, opened 2019-05-16
#38 - Fixed server side rendering from function-aws-api-proxy

jschwartz73:


graemerocher:
Thanks for the PR! This change also alters the support Micronaut version from 1.0.x to 1.1.x we would need to change the version in `build.gradle` to a new minor version to reflect that https://github.com/micronaut-projects/micronaut-aws/blob/master/gradle.properties#L1

From 1.1.2 snapshot to 1.2 snapshot

jschwartz73:
@graemerocher - The only reason I used 1.1.1 is because I'm a FreeMarker user and FM was introduced in 1.1.1. 

How about if I switch the Micronaut version back to 1.0.4 and use a different render engine for this PR?

graemerocher:
Sounds good. Thanks

PR #47, opened 2019-06-14
Fixed mixing of greeting and echo

cero-t:


cero-t:
oops, method name should be 'echo' not 'sayHello'.

PR #85, opened 2019-10-13
API Proxy lambda handle static resource routing

### LINKED ISSUE #67: MicronautLambdaContainerHandler does not proxy static routes
If you set up your fatJar with some static resources and configure the app like so....

```yaml
micronaut:
  security:
    enabled: true
    intercept-url-map:
      -
        pattern: /**
        http-method: GET
        access:
          - isAnonymous()
  router:
    static-resources:
      default:
        enabled: true
        mapping: "/**"
        paths: "classpath:public"

```

Requests like `GET /statics/icon/favicon.ico` are not proxied.  Therefore gateway just returns a 404. 

This makes it hard to deliver something like an SPA out of your jar.

viniciuslrangel:
Fixes #67


graemerocher:
Thanks for the contribution. Made some changes to improve error handling but otherwise ok

PR #158, opened 2020-01-03
Fix for AWS Lambda Container handler to support media type ALL for gr…

ShashankMishra:
…aphQL

Fix for AWS Lambda Container handler to support media type ALL for graphQL requests, Removed unused import.

Already discussed it over graphQL thread.


ShashankMishra:
@graemerocher Please merge to support graphQL requests.

PR #195, opened 2020-02-14
Using a different ObjectMapper if configured to do so

### LINKED ISSUE #186: Changing jackson config causes weird behavior of Lambda request handling
I found a problem with the whole Lambda handling, when changing the configuration of the default `ObjectMapper` in Micronaut:
```
jackson:
  bean-introspection-module: true
  propertyNamingStrategy: SNAKE_CASE
  serializationInclusion: NON_ABSENT
  dateFormat: yyyy-MM-dd'T'HH:mm:ss:SSSZ
  timeZone: CET
```

This is because the `MicronautLambdaContainerHandler` uses the global `ObjectMapper`. I would expect, that the framework would use a seperate one, which is configured to just met the Lambda specifications or at least that the properties of classes used for Lambda API (like `AwsProxyRequest`) are annotated with `@JsonProperty` etc.

Im using Micronaut `1.2.9`

alvarosanchez:
Fixes #186

### DIFF
--- a/aws-common/build.gradle
+++ b/aws-common/build.gradle
@@ -2,7 +2,7 @@ group "io.micronaut.configuration"
 
 dependencies {
     annotationProcessor "io.micronaut:micronaut-inject-java"
-    annotationProcessor "io.micronaut.docs:micronaut-docs-asciidoc-config-props:$micronautDocsVersion"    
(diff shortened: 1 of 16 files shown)


graemerocher:
Looks ok, why did you change the docs version to be hard coded? Is it so the bot maintains it?

alvarosanchez:
Correct.

Issue #199, opened 2020-02-19
Improve warm start time for AWS Lambda


eugener:
AWS Lambda limits resources (CPU and memory) only during `handleRequest` execution.  
It might be beneficial to move application context creation, context bean registration and injection of request handler instance out of `handleRequest` method  ( possibly into constructor ) and  leave only function execution in it. 

This seems to be already happening within `MicronautRequestStreamHandler` but not in `MicronautRequestHandler`

I can create a PR if it makes sense.

graemerocher:
PRs welcome

PR #200, opened 2020-02-19
Improve AWS Lambda execution time for MicronautRequestHandler

eugener:
Improve AWS Lambda execution time on `MicronautRequestHandler` by moving application context building and handler instance injection into default constructor, which is given more CPU and memory resources comparing to `handleRequest` method. Fixes issue #199

### DIFF
--- a/function-aws/src/main/java/io/micronaut/function/aws/MicronautRequestHandler.java
+++ b/function-aws/src/main/java/io/micronaut/function/aws/MicronautRequestHandler.java
@@ -28,8 +28,6 @@ import io.micronaut.function.executor.AbstractFunctionExecutor;
 import javax.annotation.Nonnull;
 import java.util.Optional;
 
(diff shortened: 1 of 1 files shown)


function-aws/src/main/java/io/micronaut/function/aws/MicronautRequestHandler.java
```diff
@@ -43,19 +43,24 @@
...
+
         if (context != null) {
             registerContextBeans(context, applicationContext);
         }
 
         if (!inputType.isInstance(input)) {
             input = convertInput(input);
         }
-        return applicationContext.inject(this).execute(input);
+        return injectedInstance.execute(input);
```
graemerocher: Only change is I don't think the extra field and `injectedInstance` is necessary, since you can just refer to `this`

eugener: Done!

graemerocher:
Thanks for the contribution

PR #225, opened 2020-04-01
More details to release process

sdelamo:


graemerocher:
@alvarosanchez can confirm but I don't believe those steps are necessary. All you have to do is publish the release from the releases page

alvarosanchez:
Indeed they are not necessary. I will also reverts the commits you manually made since that is handled by the release process

graemerocher:
btw the latest release run failed https://github.com/micronaut-projects/micronaut-aws/actions/runs/68143913 

The docs didn't publish. Seems central sync failed

alvarosanchez:
Yeah, we will undo stuff and retry

PR #226, opened 2020-04-01
Extract AwsLambdaRuntimeApi

sdelamo:


graemerocher:
Please fix the checkstyle errors first

sdelamo:
sorry, I forced pushed the checkstyle fix

graemerocher:
So the only thing that confuses me with this is why an interface with static methods? Wouldn't it be better to make these default methods?

sdelamo:
I've changed the methods to be `default` instead of `static`.

function-aws-custom-runtime/src/main/java/io/micronaut/function/aws/runtime/MicronautLambdaRuntime.java
```diff
@@ -48,17 +45,8 @@
...
-     */
-    @Deprecated
-    public static final String HEADER_RUNTIME_AWS_REQUEST_ID = LambdaRuntimeInvocationResponseHeaders.LAMBDA_RUNTIME_AWS_REQUEST_ID;
-
-    public static final UriTemplate INVOCATION_TEMPLATE = UriTemplate.of("/2018-06-01/runtime/invocation/{requestId}/response");
-    public static final UriTemplate ERROR_TEMPLATE = UriTemplate.of("/2018-06-01/runtime/invocation/$requestId/error");
-    public static final String NEXT_INVOCATION_URI = "/2018-06-01/runtime/invocation/next";
-    public static final String INIT_ERROR_URI = "/2018-06-01/runtime/init/error";
-
+    protected AwsLambdaRuntimeApi awslambdaRuntimeApi = new AwsLambdaRuntimeApi() { };
```
graemerocher, line 48: Can `MicronautLambdaRuntime` not just implement `AwsLambdaRuntimeApi` instead of creating an anonymous instance?

sdelamo: done.

PR #269, opened 2020-05-01
Support for binary requests for the API Gateway

musketyr:
The current implementation only supports JSON inputs as all other content types must be sent in binary format. This PR adds the ability to handle binary requests.

The only problem with this PR is that it seems to me that there is a lack of converters from `byte[]` to other objects. Even automatic conversion from `byte[]` to `String` does not work. But this must be addressed in core libraries, not the AWS one.


graemerocher:
Thanks for the contribution!

PR #285, opened 2020-05-14
GraalVM configuration files for AWS Lambda Custom Runtime

alvarosanchez:
This makes AWS SDK v2 with the Apache HTTP client to work inside a native image.

The Netty classes that I had to include I'm not sure if are coming from Micronaut (and hence we should move that config to core) or from AWS SDK's Netty client implementation.

### DIFF
--- a/function-aws-custom-runtime/src/main/resources/META-INF/native-image/io/micronaut/function/aws/runtime/native-image.properties
+++ b/function-aws-custom-runtime/src/main/resources/META-INF/native-image/io/micronaut/function/aws/runtime/native-image.properties
@@ -1,0 +1,7 @@
+Args = --report-unsupported-elements-at-runtime \
+       --initialize-at-run-time=io.netty.channel.epoll.Epoll,io.netty.channel.epoll.Native,io.netty.channel.epoll.EpollEventLoop,io.netty.channel.epoll.EpollEventArray,io.netty.channel.unix.Errors,io.netty.channel.unix.IovArray,io.netty.channel.unix.Socket,org.apache.commons.logging.LogFactory \
+       --initialize-at-build-time=org.apache.http.pool.ConnPoolControl,org.apache.http.HttpClientConnection,org.apache.http.conn.routing.HttpRoute,org.apache.http.conn.HttpClientConnectionManager,org.apache.http.protocol.HttpContext,org.apache.http.conn.ConnectionRequest,org.apache.http.client.config.RequestConfig,org.apache.http.client.config.RequestConfig$Builder \
+       -H:ReflectionConfigurationResources=${.}/reflection-config.json \
+       -H:DynamicProxyConfigurationResources=${.}/dynamic-proxy-config.json \
+       -H:+TraceClassInitialization \
+       -H:+ReportExceptionStackTraces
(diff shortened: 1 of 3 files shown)


function-aws-custom-runtime/src/main/resources/META-INF/native-image/io/micronaut/function/aws/runtime/native-image.properties
```diff
@@ -0,0 +1,7 @@
+Args = --report-unsupported-elements-at-runtime \
```
graemerocher, line 1: This flag needs to be removed

function-aws-custom-runtime/src/main/resources/META-INF/native-image/io/micronaut/function/aws/runtime/native-image.properties
```diff
@@ -0,0 +1,7 @@
...
+       --initialize-at-run-time=io.netty.channel.epoll.Epoll,io.netty.channel.epoll.Native,io.netty.channel.epoll.EpollEventLoop,io.netty.channel.epoll.EpollEventArray,io.netty.channel.unix.Errors,io.netty.channel.unix.IovArray,io.netty.channel.unix.Socket,org.apache.commons.logging.LogFactory \
+       --initialize-at-build-time=org.apache.http.pool.ConnPoolControl,org.apache.http.HttpClientConnection,org.apache.http.conn.routing.HttpRoute,org.apache.http.conn.HttpClientConnectionManager,org.apache.http.protocol.HttpContext,org.apache.http.conn.ConnectionRequest,org.apache.http.client.config.RequestConfig,org.apache.http.client.config.RequestConfig$Builder \
+       -H:ReflectionConfigurationResources=${.}/reflection-config.json \
+       -H:DynamicProxyConfigurationResources=${.}/dynamic-proxy-config.json \
+       -H:+TraceClassInitialization \
```
graemerocher, line 6: This doesn't belong here either

function-aws-custom-runtime/src/main/resources/META-INF/native-image/io/micronaut/function/aws/runtime/native-image.properties
```diff
@@ -0,0 +1,7 @@
...
+       -H:+ReportExceptionStackTraces
```
graemerocher, line 7: Same with this please remove

function-aws-custom-runtime/src/main/resources/META-INF/native-image/io/micronaut/function/aws/runtime/native-image.properties
```diff
@@ -0,0 +1,7 @@
+Args = --report-unsupported-elements-at-runtime \
+       --initialize-at-run-time=io.netty.channel.epoll.Epoll,io.netty.channel.epoll.Native,io.netty.channel.epoll.EpollEventLoop,io.netty.channel.epoll.EpollEventArray,io.netty.channel.unix.Errors,io.netty.channel.unix.IovArray,io.netty.channel.unix.Socket,org.apache.commons.logging.LogFactory \
+       --initialize-at-build-time=org.apache.http.pool.ConnPoolControl,org.apache.http.HttpClientConnection,org.apache.http.conn.routing.HttpRoute,org.apache.http.conn.HttpClientConnectionManager,org.apache.http.protocol.HttpContext,org.apache.http.conn.ConnectionRequest,org.apache.http.client.config.RequestConfig,org.apache.http.client.config.RequestConfig$Builder \
```
graemerocher, line 3: Are these classes on the classpath by default for the lambda runtime?

alvarosanchez: They are when you include the AWS SDK, eg:

```groovy
implementation platform("software.amazon.awssdk:bom:$awsSdkVersion")
implementation "software.amazon.awssdk:s3"
implementation "software.amazon.awssdk:rekognition"
```

graemerocher: Then we would need these applied to new modules since if they are not on the classpath these entries will break the custom runtime on Graal

alvarosanchez:
Closing since this needs a better organisation

PR #288, opened 2020-05-16
return sourceIp from MicronautAwsProxyRequest getRemoteAddress()

lestephane:
Without this functionality, I get an NPE in IpCheckRule when using these two features with create-app 

- aws-api-gateway
- security-jwt

### DIFF
--- a/function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautAwsProxyRequest.java
+++ b/function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautAwsProxyRequest.java
@@ -38,6 +38,7 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.ws.rs.core.SecurityContext;
 import java.io.UnsupportedEncodingException;
+import java.net.InetSocketAddress;
 import java.net.URI;
 import java.net.URLDecoder;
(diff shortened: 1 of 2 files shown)


function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautAwsProxyRequest.java
```diff
@@ -193,6 +194,14 @@ public URI getUri() {
         return URI.create(getScheme() + "://" + hostHeader + path);
     }
 
+    @Nonnull
+    @Override
+    public InetSocketAddress getRemoteAddress() {
+        final String sourceIp = this.awsProxyRequest.getRequestContext().getIdentity().getSourceIp();
```
graemerocher, line 200: Can you add null checks here because it seems there is a possibility that `getRequestContext()` and/or `getIdentity()` could be `null`

If the source IP cannot be resolved you can fall back to the super implementation

graemerocher:
Applied manually with changes https://github.com/micronaut-projects/micronaut-aws/commit/284b38faccc757f08b21d8b2305b909627c9994b

Thanks

lestephane:
@graemerocher thx, I did not know what to provide for the port, by since InetSocketAddress is in use, I had to make something. The remote port information is not available in lambda.

Issue #52, opened 2019-07-15
MicronautAwsProxyExceptionHandler giving 500 in case of JSON Mapping exception


ShashankMishra:
We are deserialising request json dates with our custom deserialiser.
In case of invalid date format we are throwing our custom exception which ultimately gets wrapped in JSONMappingException and later on to InvalidRequestEventException.

Now since I want to send 400 BAD REQUEST in such cases but **MicronautAwsProxyExceptionHandler** is handling these exception by default and send 

> 500 Internal Server Error 

in place of 

> 400 BAD REQUEST.

We tried creating our own error handler and tried replacing **MicronautAwsProxyExceptionHandler** but that too is not working because this exception handler is directly injected in **MicronautLambdaContainerHandler**.

Flow :
1.  Request forwarded to lambda handler  from our function -

```java
    @Override
    public void handleRequest(InputStream input, OutputStream output, Context context) throws IOException {
        logger.info("Input: {}", input);
        handler.proxyStream(input, output, context);
    } 
```

2.  Gets exception generated due to json deserialisation of wrong format  of request at - **MicronautLambdaContainerHandler.java**

```java
    public void proxyStream(InputStream input, OutputStream output, Context context)
            throws IOException {

        try {
            RequestType request = readerFor(requestTypeClass).readValue(input);
            ResponseType resp = proxy(request, context);

            writerFor(responseTypeClass).writeValue(output, resp);
        } catch (JsonParseException e) {
            log.error("Error while parsing request object stream", e);
            objectMapper().writeValue(output, exceptionHandler.handle(e));
        } catch (JsonMappingException e) {
            log.error("Error while mapping object to RequestType class", e);
            objectMapper().writeValue(output, exceptionHandler.handle(e));
        } finally {
            output.flush();
            output.close();
        }
    }
```
3. MicronautAwsProxyExceptionHandler  handling exception and sending 500 in place of 400 Bad request.

```java
    @Override
    public AwsProxyResponse handle(Throwable ex) {
        LOG.error("Called exception handler for:", ex);

        // adding a print stack trace in case we have no appender or we are running inside SAM local, where need the
        // output to go to the stderr.
        ex.printStackTrace();
        **if (ex instanceof InvalidRequestEventException) {
            return new AwsProxyResponse(500, headers,** getErrorJson(INTERNAL_SERVER_ERROR));
        } else {
[...]

PR #306, opened 2020-05-29
Allow invalid JSON requests to be handled by user defined exception handlers.

magJ:
Instead of throwing a 500 series server error, allow handling of bad JSON requests,
so that we can return a more appropriate bad request error.

Fixes: #52

The attempt by PR #169 to fix issue #52, miss-understood the problem.  
JSON decoding errors are thrown before ever entering a route, thus this change moves the request body decoding logic into the handler, and makes the exception available to user-defined exception handlers.


graemerocher:
Thanks for the contribution

Issue #376, opened 2020-06-24
Support disabling System Manager Parameter Store queries by micronaut environment


sam-lewis:
The start-up performance of my AWS lambda function is negatively impacted by the number of requests to System Manager Parameter Store.

Currently paths for every active micronaut profile are queried e.g.
- `/config/application`
- `/config/myapp`
- `/config/application_EC2`
- `/config/myapp_EC2`
- `/config/application_lambda`
- `/config/myapp_lambda`
- etc etc

I would like the ability to only query paths `/config/application` and`/config/myapp`.

This could be achieved with a new property `aws.client.system-manager.parameterstore.searchActiveEnvironments` which defaults to true in order to preserve existing behaviour.

I have been testing this in a fork and can open a pull request if desirable.

graemerocher:
Seems like a good idea, PRs welcome

sam-lewis:
Thanks @graemerocher I opened a draft PR

PR #379, opened 2020-06-25
Disable System Manager Parameter Store queries by micronaut environment

sam-lewis:
PR resolves https://github.com/micronaut-projects/micronaut-aws/issues/376

### DIFF
--- a/aws-parameter-store/src/main/java/io/micronaut/discovery/aws/parameterstore/AWSParameterStoreConfigClient.java
+++ b/aws-parameter-store/src/main/java/io/micronaut/discovery/aws/parameterstore/AWSParameterStoreConfigClient.java
@@ -59,8 +59,7 @@ import java.util.concurrent.Future;
 @BootstrapContextCompatible
 public class AWSParameterStoreConfigClient implements ConfigurationClient {
 
-    private static final Logger LOG = LoggerFactory.getLogger(AWSParameterStoreConfiguration.class);
-    private final AWSClientConfiguration awsConfiguration;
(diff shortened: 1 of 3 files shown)


sam-lewis:
I am not very happy with the test, any suggestions on the best approach to creating `AWSParameterStoreConfigClient` with different configurations in different test scenarios would be appreciated.

Also, I did a tiny but of refactoring but that could be done as a separate MR:
- Correct a logger class
- Remove an unused variable
- Fix javadoc variable substitutions
- Remove semicolons from groovy file

graemerocher:
The tests for this integration are complex yes, suggestions on how we can improve this welcome.

sam-lewis:
Integration tests with [Localstack](https://github.com/localstack/localstack) and [Testcontainers](https://www.testcontainers.org/modules/localstack/) could be nice because the mocking could be removed.
That would have disadvantages though:
- Build dependency on docker
- Much slower tests due to image download and container start-up
Is this something you would consider?

Alternatively a bit of refactoring to extract common methods for mock interactions and a common method to create `AWSParameterStoreConfigClient` with different configuration properties could be an improvement.

alvarosanchez:
@sam-lewis we use Testcontainers in quite a few Micronaut modules, so I don't see any problem using it for some specific tests like this one.

If you are willing to contribute that, go ahead. That said, in my opinion, this PR is good to go. If you agree, we can merge it as is, and have the Localstack+Testcontainers integration in a separate PR. Thoughts?

alvarosanchez (approved):

sam-lewis:
@alvarosanchez I think that Localstack+Testcontainers integration would be better as a separate PR so it would be good to merge this.

PR #499, opened 2020-09-10
refactor: Split proxy-test into beans

sdelamo:
@graemerocher I have split the proxy-test server into multiple beans. 

The goal is to make code a litter easier to process and also to provide users the ability to override certain beans in their test.  

toughts?


graemerocher:
I don't have any problem with these changes as such, but the the API here is not meant to be something users interact with. The original class is annotated with `@Internal`. The only thing useful I see here is `ContextProvider` which maybe the user could want to extend.

sdelamo:
Or maybe they could override the request or response adapter if we don't have a scenario properly covered. Are you ok with me merging this then?

graemerocher:
Sure

Issue #605, opened 2020-10-14
Using DynamoDB client (with URLConnection client) results in an error


ducampbe:
### Steps to Reproduce

1. Create a new project from the micronaut-cli with support for AWS SDK v2 & AWS Lambda
2. Add the following to the build.gradle
```
 implementation("software.amazon.awssdk:dynamodb") {
        exclude group: "software.amazon.awssdk", module: "apache-client"
        exclude group: "software.amazon.awssdk", module: "netty-nio-client"
}
implementation('software.amazon.awssdk:url-connection-client')
```
3. In one of the unit tests, attempt to inject `DynamoDBClient` into the class and execute the unit test

### Expected Behaviour

The unit test should complete successfully with no errors and the class should contain an instance of DynamoDBClient initialised with the synchronous client (URLConnection)

### Actual Behaviour

The unit test fails with an error complaining that `No bean of type [software.amazon.awssdk.http.async.SdkAsyncHttpClient] exists`

### Fix

It appears the async factory methods in DynamoDBClientFactory are missing the `@Requires(beans = SdkAsyncHttpClient.class)` annotation which appears in the other AWS factory classes (S3, SNS, etc.). PR to follow.

### Environment Information

- **Operating System**: `Linux desktop-i3 5.8.11-1-MANJARO #1 SMP PREEMPT Wed Sep 23 14:35:40 UTC 2020 x86_64 GNU/Linux` 
- **Micronaut Version:** `2.1.1`
- **JDK Version:** `openjdk 11.0.8 2020-07-14`

### Example Application

- https://github.com/ducampbe/micronaut-aws-url-conn-issue (Java 11 / Gradle)

PR #606, opened 2020-10-14
Fixes issue of attempting to inject SdkAsyncHttpClient when using URLConnection client...

ducampbe:
...and explicitly excluding the async HTTP Clients. Addresses #605


alvarosanchez (approved):
Thanks for the contribution!

PR #613, opened 2020-10-16
Initialize eagerly singletons and configuration

ilopmar:
Add missing eager initialization of singletons and configuration in handler used as entry point for apps.

This fix needs https://github.com/micronaut-projects/micronaut-core/issues/4316 which will be included in Micronaut 2.1.2.

In my test application the cold start went from 12 seconds to 9 with this change.


graemerocher:
Failing due to https://github.com/micronaut-projects/micronaut-core/issues/4316

ilopmar:
Yeah, exactly. We need to wait for Micronaut 2.1.2 before merge this.

PR #626, opened 2020-10-20
feat: adds mdc keys for context entries and xray

sdelamo:
This allows to have log pattern with context related info: 

```
2020-10-20 11:02:25 0aedcbd6-50ea-4deb-bc8e-6de34c6cefae l337-java11 $LATEST arn:aws:lambda:eu-west-1:040181416768:function:l337-java11 512  14998 1-5f8ec3c1-07e6491a3e5d006144559d2f INFO  c.m.t.p.TelegramParser - text parsed: are you a robot?
```


graemerocher:
Seems like this is a lot of code and changes for what could simply be a `protected` method added to `MicronautRequestHandler` to set these values in MDC which the user could override to customize or disable or am I missing something?

sdelamo:
I could do what you suggest. I thought it was cleaner to have it on its own set of classes. 

Some context. I've been checking [AWS Lambda Powertools](https://github.com/awslabs/aws-lambda-powertools-java). It has some functionality to do logging but it is coupled to Log4j. They log a JSON object with fields similar to what I am putting into MDC. One on the fields of the JSON Object is the log message. I think having the MDC logic on its own class could help us to do a generic JSON log approach in the future if we want to go that route. 

Thus, if you think it is better to just have two methods in `MicronautRequestHandler` I can do that.

graemerocher:
Given this code is performance sensitive in terms of cold starts and execution I think the less overhead the better

sdelamo:
> Given this code is performance sensitive in terms of cold starts and execution I think the less overhead the better

Done the changes you requested @graemerocher

PR #679, opened 2020-11-03
Configurable AWS Parameter Store queries

### LINKED ISSUE #657: Parameter Store queries could use some configurability
As stated in the documentation [here](https://docs.micronaut.io/latest/guide/index.html#distributedConfigurationAwsParameterStore), the `AWSParameterStoreConfigClient` will send queries for a number of paths computed from the application name and the active environments, following a fixed format. In one of my projects, an application needed to access values written by a configuration publishing pipeline on a path prefix over which I had no control. Playing with the `rootHierarchyPath` can help but in this case `/rootHierarchyPath/application` was off-limits and the client crashed with an `AccessDeniedException`. I had to override the entire client to get things to work and integrate properly with the `PropertySource` machinery. I am sure there are many other situations where more flexibility would be needed.

Maybe we can extract the path calculation in a dedicated component with an exposed interface, and provide users with a way to configure the target paths as they please by implementing it. The default implementation would preserve the current behaviour. What do you think ? I am ready to make a PR.

ttzn:
This PR resolves #657 by adding the ability to declare an `AWSParameterQueryResolver` in cases where the default `/config/application` and `/config/project_name` base paths are undesirable. A package-protected default implementation is provided that preserves the current behaviour.

I also took the liberty to rewrite the very long `getPropertySources()` method in a more pipeline-like fashion, with transformations encapsulated in their own methods.

There is some complexity that is due to the fact that the current implementation uses both `GetParameters` and `GetParametersByPath` AWS API calls to retrieve configuration values. Unless I missed some legacy quirk, a call such as `aws ssm get-parameters --name "/config/application"` doesn't seem to make sense, as it will only return something if a parameter named `/config/application` exists, but that would result in a property with an empty string key. I am not sure that is intended (maybe a separate issue should be opened), nevertheless I added the possibility to do both styles of queries with a `name` flag, so that all current tests pass and backwards compatibility is preserved.

### DIFF
--- a/aws-parameter-store/src/main/java/io/micronaut/discovery/aws/parameterstore/AWSParameterQueryProvider.java
+++ b/aws-parameter-store/src/main/java/io/micronaut/discovery/aws/parameterstore/AWSParameterQueryProvider.java
@@ -15,6 +15,7 @@
  */
 package io.micronaut.discovery.aws.parameterstore;
 
+import edu.umd.cs.findbugs.annotations.NonNull;
 import io.micronaut.context.annotation.DefaultImplementation;
 import io.micronaut.context.env.Environment;
 
@@ -37,9 +38,19 @@ import java.util.Optional;
  * }}
  * </pre>
  *
+ * @author ttzn
  * @since 2.1.2
  */
 @DefaultImplementation(DefaultParameterQueryProvider.class)
 public interface AWSParameterQueryProvider {
-    List<ParameterQuery> getParameterQueries(Environment environment, Optional<String> serviceId, AWSParameterStoreConfiguration configuration);
+    /**
+     * @param environment the current application environment
+     * @param serviceId the service ID or application name, if applicable
+     * @param configuration the parameter store configuration
+     * @return a list of {@link ParameterQuery} that will be used to configure calls
+     * to the Parameter Store
+     */
+    @NonNull List<ParameterQuery> getParameterQueries(@NonNull Environment environment,
+                                                      @NonNull Optional<String> serviceId,
+                                                      @NonNull AWSParameterStoreConfiguration configuration);
 }
--- a/aws-parameter-store/src/main/java/io/micronaut/discovery/aws/parameterstore/AWSParameterQueryProvider.java
+++ b/aws-parameter-store/src/main/java/io/micronaut/discovery/aws/parameterstore/AWSParameterQueryProvider.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2017-2020 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


aws-parameter-store/src/main/java/io/micronaut/discovery/aws/parameterstore/DefaultParameterQueryProvider.java
```diff
@@ -0,0 +1,69 @@
...
+import io.micronaut.context.env.EnvironmentPropertySource;
+
+import javax.inject.Singleton;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+
+@Singleton
+class DefaultParameterQueryProvider implements AWSParameterQueryProvider {
```
graemerocher, line 28: Declare final, add javadoc with `@author` and `@since` tags and add `@Internal` annotation

aws-parameter-store/src/main/java/io/micronaut/discovery/aws/parameterstore/AWSParameterQueryProvider.java
```diff
@@ -0,0 +1,45 @@
...
+ * ...
+ *
+ * }}
+ * </pre>
+ *
+ * @since 2.1.2
+ */
+@DefaultImplementation(DefaultParameterQueryProvider.class)
+public interface AWSParameterQueryProvider {
+    List<ParameterQuery> getParameterQueries(Environment environment, Optional<String> serviceId, AWSParameterStoreConfiguration configuration);
```
graemerocher, line 44: Add javadoc to the method

aws-parameter-store/src/main/java/io/micronaut/discovery/aws/parameterstore/AWSParameterQueryProvider.java
```diff
@@ -0,0 +1,45 @@
...
+ * {@code
+ * @Replaces(AWSParameterQueryProvider.class)
+ * public class CustomParameterQueryProvider implements AWSParameterQueryProvider {
+ *
+ * ...
+ *
+ * }}
+ * </pre>
+ *
+ * @since 2.1.2
```
graemerocher, line 40: Add `@author` tags

aws-parameter-store/src/main/java/io/micronaut/discovery/aws/parameterstore/AWSParameterQueryProvider.java
graemerocher: Add `@NonNull` and `@Nullable` to method arguments to describe which parameters are null and whether the return value can be null

graemerocher (changes requested):
Thanks for the contribution! Just a few changes

aws-parameter-store/src/main/java/io/micronaut/discovery/aws/parameterstore/AWSParameterQueryProvider.java
ttzn: None of those are nullable, should I add `@NonNull` to all of them ? From the rest of the code it looks like it's implied

ttzn: Added them anyway, can't hurt

graemerocher (approved):

PR #748, opened 2020-11-23
Inject preconstructed ApplicationContext in Lambda handler

ttzn:
This helps solving https://github.com/micronaut-projects/micronaut-test/issues/330. It basically allows initializing `MicronautRequestHandler`s with the app context from a `@MicronautTest` in this fashion:

```kotlin
@MicronautTest(contextBuilder = [LambdaApplicationContextBuilder::class])
class RequestHandlerTest {
    @Inject
    lateinit var context: ApplicationContext

    @Test
    fun testHandler() {
        val sampleRequestHandler = SampleRequestHandler(context)
        ...
```

This is probably still too much work on the user's part. Before going further I'd like this to be validated first.

### DIFF
--- a/function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautLambdaContainerHandler.java
+++ b/function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautLambdaContainerHandler.java
@@ -28,14 +28,13 @@ import com.fasterxml.jackson.databind.ObjectWriter;
 import io.micronaut.context.ApplicationContext;
 import io.micronaut.context.ApplicationContextBuilder;
 import io.micronaut.context.ApplicationContextProvider;
-import io.micronaut.context.env.Environment;
 import io.micronaut.core.annotation.AnnotationMetadata;
 import io.micronaut.core.annotation.TypeHint;
 import io.micronaut.core.async.publisher.Publishers;
 import io.micronaut.core.type.Argument;
 import io.micronaut.core.type.TypeVariableResolver;
 import io.micronaut.core.util.ArgumentUtils;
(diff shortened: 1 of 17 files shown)


graemerocher:
Looks like a good start. It would probably be work creating a stereotype annotations like:

```java
@MicronautTest(contextBuilder = [LambdaApplicationContextBuilder::class])
@interface MicronautLambdaTest {

}
```

It could also maybe provide injection of the handler so the user doesn't have to do that by adding additional extensions .

ttzn:
@graemerocher, a dedicated annotation is what I had in mind. However since it's not possible to "redirect" parameters from an annotation to a meta-annotation, I'll have to basically copy-paste the whole `@MicronautTest` body so that the user can still customize the application context as they please.

As for injecting handlers, the most user-friendly way would be to annotate them with `@Singleton`, but when I tried that I ran into https://github.com/micronaut-projects/micronaut-core/issues/4586 (I made an MR for that).

ttzn:
@graemerocher I added a `MicronautLambdaTest` annotation with a supporting JUnit 5 extension. Two points to consider:

* I added a new `micronaut-function-aws-test` subproject to isolate this feature;
* there is quite a bit of code that is replicated from `MicronautJunit5Extension`, a separate MR in `micronaut-test` would allow optimal code reuse if you agree with these changes.

Also maybe I should rename this branch as it has outgrown its initial purpose.

graemerocher:
Thanks, we probably need documentation for the new module and will need to bump the minor version of the project before merging this PR

ttzn:
I added documentation. It would be best if https://github.com/micronaut-projects/micronaut-test/pull/336 were merged first because it would allow adding the `contextBuilder` option and thus feature-parity with `@MicronautTest`.

graemerocher:
Thanks for the contribution

PR #880, opened 2021-01-11
Use correct discovery client dependency

graemerocher:


graemerocher:
Relates to https://github.com/micronaut-projects/micronaut-core/issues/4295

alvarosanchez (approved):

Issue #881, opened 2021-01-15
MicronautLambdaRuntime swallows exception root cause


RoadRunner120485:
Sorry for not filling the issue template but hopefully it is not necessary for my concern. 

I'm running a lambda, with Micronaut, GraalVM and a custom Runtime as supposed in the corresponding example. I've now updated form an Micronaut 1.3.x to 2.2.3 and apart from the fact that the update did not took much effort, in the end my lamda crashed with just the following exception:
```
io.micronaut.context.exceptions.ConfigurationException: Exception thrown instantiating MicronautLambdaRuntimeHandler
        at io.micronaut.function.aws.runtime.MicronautLambdaRuntime.createRequestHandler(MicronautLambdaRuntime.java:40)
        at io.micronaut.function.aws.runtime.AbstractMicronautLambdaRuntime.createHandler(AbstractMicronautLambdaRuntime.java:192)
        at io.micronaut.function.aws.runtime.AbstractMicronautLambdaRuntime.startRuntimeApiEventLoop(AbstractMicronautLambdaRuntime.java:313)
        at io.micronaut.function.aws.runtime.AbstractMicronautLambdaRuntime.run(AbstractMicronautLambdaRuntime.java:129)
        at io.micronaut.function.aws.runtime.MicronautLambdaRuntime.main(MicronautLambdaRuntime.java:50)
Request loop failed with: Exception thrown instantiating MicronautLambdaRuntimeHandler
```

And nothing else. I found out, that `MicronautLambdaRuntime` catches any exception which raises and just throws a `ConfigurationException` without passing the root cause to it. So I was not able to figure out what the root cause was. I managed to get it working now by patching Micronaut, build my project using a patched snapshot of Micronaut and executing my lambda locally with SAM local, but it was not very comfortable...

So long story short: I'd like to suggest an improvement for `MicronautLambdaRuntime`:
```
    @Override
    protected RequestHandler<AwsProxyRequest, AwsProxyResponse> createRequestHandler(String... args) {
        try {
            return new MicronautLambdaHandler(createApplicationContextBuilderWithArgs(args));
        } catch (ContainerInitializationException e) {
            throw new ConfigurationException("Exception thrown instantiating MicronautLambdaRuntimeHandler", e); // pass root cause here!
        }
    }
```

TY guys.

graemerocher:
PRs welcome!

RoadRunner120485:
Great :)

PR #882, opened 2021-01-15
Append cause to thrown exception.

RoadRunner120485:
Here the PR as suggested in #881


graemerocher:
Thanks for the contribution!

PR #916, opened 2021-02-10
logging fixes

burtbeckwith:
don't use unnecessary log.isEnabled() checks when using SLF4J, use placeholder replacement
avoid unnecessary toString() when logging
reworked LambdaRuntime logging in AbstractMicronautLambdaRuntime to avoid unnecessary string concatenation

### DIFF
--- a/aws-alexa-httpserver/src/main/java/io/micronaut/aws/alexa/httpserver/controllers/SkillController.java
+++ b/aws-alexa-httpserver/src/main/java/io/micronaut/aws/alexa/httpserver/controllers/SkillController.java
@@ -106,10 +106,7 @@ public class SkillController {
             }
 
         } catch (IOException e) {
-
-            if (LOG.isErrorEnabled()) {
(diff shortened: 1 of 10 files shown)


sdelamo:
@burtbeckwith it looks great. 

We are using `isEnabled` all through the framework, Are you 100% sure it is not necessary ? I think you are correct but why is it SLF4J API if it is not?

sdelamo (approved):

burtbeckwith:
See http://www.slf4j.org/faq.html#logging_performance for info on why isEnabled checks are generally not necessary as long as placeholders are used. Let's discuss Tuesday.

sdelamo:
> See http://www.slf4j.org/faq.html#logging_performance for info on why isEnabled checks are generally not necessary as long as placeholders are used. Let's discuss Tuesday.

You are 100% right. We should not us the if statement when using parametrised. It is also ugly code which has always bothered me.

PR #958, opened 2021-03-08
fix failure when body was already decoded

### LINKED ISSUE #136: No CORS headers returned (again) if exception is raised with API Gateway Proxy + Lambda, GraalVM native image
Micronaut version: 1.2.6

Setup:
I have an API Gateway Proxy + Lambda app created with `--features aws-api-gateway-graal`, but I'm temporarily using the Java8 runtime due to GraalVM's issue with the MySQL driver.

The issue I'm having is that there are no CORS headers returned when exceptions are raised, for example, HTTP input params validation errors.

Basically the same as with this:
https://github.com/micronaut-projects/micronaut-core/issues/1471

Also, I cannot use a custom exception handler due to this issue:
https://github.com/micronaut-projects/micronaut-aws/issues/125

In this case there were also no CORS headers returned.

### LINKED ISSUE #208: CORS headers should be present for deserialization failure
Setup: API Gateway Proxy + Lambda

CORS headers should be present after a deserialization failure.

See the test case below. 

```
class BugSpec extends Specification {

    @Shared @AutoCleanup MicronautLambdaContainerHandler handler = new MicronautLambdaContainerHandler(
                ApplicationContext.build().properties([
                        'micronaut.server.cors.enabled': true,
                        'micronaut.server.cors.configurations.web.allowedOrigins': ['http://localhost:8080']
                ])
    )
    @Shared Context lambdaContext = new MockLambdaContext()

    void "CORS headers present after deserialization failure"() {
        given:
        AwsProxyRequestBuilder builder = new AwsProxyRequestBuilder('/response-body/pojo', HttpMethod.POST.toString())
        builder.header(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON)
        builder.body('{"x":10,"y":"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa20"}')

        when:
        JsonMediaTypeCodec codec = handler.getApplicationContext().getBean(JsonMediaTypeCodec)
        print(codec)
        def response = handler.proxy(builder.build(), lambdaContext)

        then:
        response.multiValueHeaders.getFirst(HttpHeaders.ACCESS_CONTROL_ALLOW_ORIGIN) == 'http://localhost:8080'
    }

    @Controller('/response-body')
    static class BodyController {

        @Post(uri = "/pojo")
        @Status(HttpStatus.CREATED)
        Point post(@Body Point data) {
            return data
        }
    }

    @Canonical
    static class Point {
        Integer x,y
    }
}
```

### LINKED ISSUE #761: "Error decoding JSON stream" in @Error methods
Thanks for reporting an issue, please review the task list below before submitting the
issue. Your issue report will be closed if the issue is incomplete and the below tasks not completed.

NOTE: If you are unsure about something and the issue is more of a question a better place to ask questions is on Stack Overflow (https://stackoverflow.com/tags/micronaut) or Gitter (https://gitter.im/micronautfw/). DO NOT use the issue tracker to ask questions.

### Task List

- [x] Steps to reproduce provided
- [x] Stacktrace (if present) provided
- [x] Example that reproduces the problem uploaded to Github
- [x] Full description of the issue provided (see below)

### Steps to Reproduce

1. Create a Micronaut application w/ Lambda support.
1. Create a controller and a method that takes an argument.
2. Define an `@Error` handler.
3. Throw an exception from your method.

### Expected Behaviour

Exception-handling generates the response that gets returned to the user. Instead a 500 is thrown and InternalServerError is returned in a JSON response.

### Actual Behaviour

A rather large Jackson-related stacktrace is returned; see attached.

### Environment Information

- **Operating System**: Fedora 33
- **Micronaut Version:** 2.2.0
- **JDK Version:** OpenJDK 64-Bit Server VM 18.9 (build 11.0.9+11, mixed mode, sharing)

### Example Application

* https://github.com/nanodeath/micronaut-json-error-bug
* [Controller](https://github.com/nanodeath/micronaut-json-error-bug/blob/main/src/main/java/com/example/BookController.java)

pgressa:
Fixes #761 

The root cause is that when the [ErrorRoute](https://github.com/micronaut-projects/micronaut-aws/blob/master/function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautLambdaContainerHandler.java#L357) is executed, then the decodeBody is called [once again](https://github.com/micronaut-projects/micronaut-aws/blob/f480262e5bbffd0a5e63c7cadfc74203859ea244/function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautLambdaContainerHandler.java#L511), what fails, because the request doesn't contain json node, but already decoded object. 

Fixes #208  
Fixes #136

### DIFF
--- a/function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautAwsProxyRequest.java
+++ b/function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautAwsProxyRequest.java
@@ -379,6 +379,13 @@ public class MicronautAwsProxyRequest<T> implements HttpRequest<T> {
         this.decodedBody = decodedBody;
     }
 
+    /**
+     * @return true if body was already decoded, false otherwise
+     */
+    boolean isBodyDecoded(){
+        return decodedBody != null;
+    }
+
     private String getScheme() {
         // if we don't have any headers to deduce the value we assume HTTPS - API Gateway's default
         if (awsProxyRequest.getMultiValueHeaders() == null) {
--- a/function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautAwsProxyRequest.java
+++ b/function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautAwsProxyRequest.java
@@ -382,7 +382,7 @@ public class MicronautAwsProxyRequest<T> implements HttpRequest<T> {
     /**
      * @return true if body was already decoded, false otherwise
      */
-    boolean isBodyDecoded(){
+    boolean isBodyDecoded() {
         return decodedBody != null;
     }
 
--- a/function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautAwsProxyRequest.java
+++ b/function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautAwsProxyRequest.java
@@ -382,7 +382,7 @@ public class MicronautAwsProxyRequest<T> implements HttpRequest<T> {
     /**
      * @return true if body was already decoded, false otherwise
      */
-    boolean isBodyDecoded() {
+    public boolean isBodyDecoded() {
         return decodedBody != null;
     }
 
--- a/function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautLambdaContainerHandler.java
+++ b/function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautLambdaContainerHandler.java
@@ -352,10 +352,15 @@ public final class MicronautLambdaContainerHandler
                                     );
                                 }
                             } else if (errorHandler instanceof MethodBasedRouteMatch) {
-                                return executeRoute(
+                                final Publisher<? extends MutableHttpResponse<?>> errorPublisher = executeRoute(
                                         containerRequest,
                                         containerResponse,
                                         (MethodBasedRouteMatch) errorHandler
+                                );
+
+                                return filterPublisher(
+                                        requestReference,
+                                        errorPublisher
                                 );
                             }
                             return Flowable.error(throwable);
--- a/function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautLambdaContainerHandler.java
+++ b/function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautLambdaContainerHandler.java
@@ -49,10 +49,7 @@ import io.micronaut.http.server.exceptions.ExceptionHandler;
 import io.micronaut.http.server.types.files.StreamedFile;
 import io.micronaut.inject.qualifiers.Qualifiers;
 import io.micronaut.jackson.codec.JsonMediaTypeCodec;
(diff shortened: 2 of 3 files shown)


graemerocher (approved):

pgressa:
@graemerocher added one more commit that fixes another bug https://github.com/micronaut-projects/micronaut-aws/pull/958/commits/ec2d4e5e8ada7393953acfe54b8872f7f8d572d8

function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautAwsProxyRequest.java
```diff
@@ -379,6 +379,13 @@ void setDecodedBody(T decodedBody) {
         this.decodedBody = decodedBody;
     }
 
+    /**
+     * @return true if body was already decoded, false otherwise
+     */
+    boolean isBodyDecoded() {
```
sdelamo, line 385: should this method be `public` or is it scoped `package-private` on purpouse.

function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautLambdaContainerHandler.java
```diff
@@ -49,10 +49,7 @@
 import io.micronaut.http.server.types.files.StreamedFile;
 import io.micronaut.inject.qualifiers.Qualifiers;
 import io.micronaut.jackson.codec.JsonMediaTypeCodec;
-import io.micronaut.web.router.MethodBasedRouteMatch;
-import io.micronaut.web.router.RouteMatch;
-import io.micronaut.web.router.Router;
-import io.micronaut.web.router.UriRouteMatch;
+import io.micronaut.web.router.*;
```
sdelamo, line 52: Please, don't use wildcard imports. 

It is easy [disable wildcard imports in IntelliJ IDEA](https://www.jetbrains.com/help/idea/creating-and-optimizing-imports.html#disable-wildcard-imports).

sdelamo (changes requested):

function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautLambdaContainerHandler.java
sdelamo: No, it is not a new rule. I just assumed we were doing that. Are you against avoiding wilcard imports @graemerocher ?

graemerocher: A rule that is not in the check style is not a rule. There are wildcards used all over the codebase right now. 

If we all agree that this should be a rule then the check style needs updating to reflect that.

sdelamo: Ok, I will bring this conversation to the group discussions. @pgressa do as you prefer regarding the wildcard imports.

sdelamo (approved):

darrenoc3:
Thanks for this @pgressa , really useful fix for us. Prevents misleading CORS errors!

PR #963, opened 2021-03-10
Add bom

### LINKED ISSUE #325: Create a project BOM

pgressa:
Fixes #325

### DIFF
--- a/aws-bom/build.gradle
+++ b/aws-bom/build.gradle
@@ -1,0 +1,15 @@
+javaPlatform {
+    allowDependencies()
+}
+dependencies {
+    constraints {
+        for (Project p : rootProject.subprojects) {
+            if (p.name.endsWith("-bom") ||  p.name.contains('example')) {
+            	continue	
+        	} 
+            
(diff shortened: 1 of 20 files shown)


aws-bom/build.gradle
```diff
@@ -0,0 +1,15 @@
+javaPlatform {
+    allowDependencies()
+}
+dependencies {
+    constraints {
```
graemerocher, line 5: Shouldn't the AWS sdk version etc. be here as well?

pgressa: it is present in the [sdk module](https://github.com/micronaut-projects/micronaut-aws/blob/master/aws-sdk-v2/build.gradle#L6) so I think that's enough and the modules are either using v1 or v2, so question is whether we want to spoil every module with both sdk boms. 

Note: The bom is not defined for v1, so I'll add it there

graemerocher (approved):

sdelamo:
I will merge this today.

sdelamo (approved):

Issue #990, opened 2021-03-18
Document correctly IntentHandler annotation


sdelamo:
Currently, `@IntentHandler` annotation is broken.

PR #993, opened 2021-03-18
doc: delete broken section about `@IntentHandler`

sdelamo:
See: https://github.com/micronaut-projects/micronaut-aws/issues/990


sdelamo:
I am deleting the docs because they are broken. I am leaving the issue open if we want to revisit it.

Issue #1006, opened 2021-03-30
Support AWS Secret Manager


sdelamo:
https://aws.amazon.com/secrets-manager/

Probably similar to our support for AWS parameter store.

tburch:
@sdelamo if you have an outline of how you envision this working, I'd be happy to help with a PR.

sdelamo:
> @sdelamo if you have an outline of how you envision this working, I'd be happy to help with a PR.

@tburch I am working on this PR: https://github.com/micronaut-projects/micronaut-aws/pull/1056

sdelamo:
@tburch I am writing a guide as well: https://github.com/micronaut-projects/micronaut-guides/pull/239

PR #1007, opened 2021-03-30
feat: add secrets manager client factory

sdelamo:
see: https://github.com/micronaut-projects/micronaut-aws/issues/1006

see: https://aws.amazon.com/secrets-manager/

### DIFF
--- a/aws-sdk-v2/build.gradle
+++ b/aws-sdk-v2/build.gradle
@@ -16,6 +16,7 @@ dependencies {
     compileOnly "software.amazon.awssdk:sns"
     compileOnly "software.amazon.awssdk:sqs"
     compileOnly "software.amazon.awssdk:ssm"
+    compileOnly 'software.amazon.awssdk:secretsmanager'
 
     // Tests
     testImplementation "software.amazon.awssdk:url-connection-client"
@@ -24,6 +25,7 @@ dependencies {
     testImplementation "software.amazon.awssdk:s3"
     testImplementation "software.amazon.awssdk:dynamodb"
     testImplementation "software.amazon.awssdk:ses"
+    testImplementation 'software.amazon.awssdk:secretsmanager'
     testImplementation "software.amazon.awssdk:sns"
     testImplementation "software.amazon.awssdk:sqs"
     testImplementation "software.amazon.awssdk:ssm"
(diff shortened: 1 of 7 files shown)


aws-sdk-v2/build.gradle
```diff
@@ -24,6 +25,7 @@ dependencies {
     testImplementation "software.amazon.awssdk:s3"
     testImplementation "software.amazon.awssdk:dynamodb"
     testImplementation "software.amazon.awssdk:ses"
+    testImplementation 'software.amazon.awssdk:secretsmanager'
```
pgressa, line 28: to stay consistent you should use double quotes  here instead of simple quotes

sdelamo: Done

pgressa (approved):
code looks ok, just add docs, as we have for every client quite generic doc page

pgressa:
@sdelamo maybe of this PR we could discuss this one https://github.com/micronaut-projects/micronaut-aws/pull/1017 that generates those factories instead of writing them manually.

sdelamo:
> @sdelamo maybe of this PR we could discuss this one #1017 that generates those factories instead of writing them manually.

@pgressa  I am going to merge this PR for now. I want to develop further the integration with secret manager for distributed configuration similar AWS Parameter store and it will need to be `@BootstrapContextCompatible`.

athkalia:
Hey @sdelamo , I am looking to load some DB credentials from AWS secrets manager to use in my lambda function that is built with Micronaut. I noticed that the guide [here](https://docs.micronaut.io/latest/guide/index.html#distributedConfigurationAwsParameterStore) only talks about the AWS parameters store. This PR seems close to what I need. Do you have any insights on whether this is possible at the moment and if so any guide for achieving what I am after? Thanks in advance! 😃

sdelamo:
>  Do you have any insights on whether this is possible at the moment and if so any guide for achieving what I am after? Thanks in advance! 😃

It is currently not possible but I want to add that capability in the short term. We have an issue for it. https://github.com/micronaut-projects/micronaut-aws/issues/1006 Please upvote it if you are interested.

athkalia:
Awesome! I will work around this for the time being, but I am definitely interested in trying it out once it's out! Make sure you include some docs in there when you are done 😄 🙏 

Do you have any sort of timelines in your mind by any chance?

PR #1017, opened 2021-03-31
Automated generate of zilion of AWS client factories

pgressa:
this has some disadvantages:

1] there are 257 bean factories generated .. I can imagine this can cause troubles for startup. So maybe instead of declaring them all, we could declare like 10 most used ones and let the users aware of this way how to micronautize the aws client what will be definitely much more efficient. 

2] it downloads all sdk for compilation, what increases the build time

3] sometimes the compilation fails b/c of heap space

Note: that list of clients was generated


sdelamo:
**I don't think we can merge this PR.** 

With these changes `aws-sdk-v2` will generate 2846 `.class` files. 

If you want to use just the simple email service to send transactional emails, nowadays you add to your application:

```
    implementation("io.micronaut.aws:micronaut-aws-sdk-v2")
    implementation("software.amazon.awssdk:ses")
```

Currently, we only have a few service client beans. We generate less than 100 classes for those. However, with this change to get a bean of type `SesClient`, he will inadvertently  we will be adding 3000 classes. Even if most of those classes don't participate in the aplication context. since they are annotated with `@Requires`. For example:`@Requires(classes = {SesClient.class, SesAsyncClient.class})`.

It would be best for the user to: 
- Create a factory in his app to generate a bean of type `SesClient` and don't add the dependency `micronaut-aws-sdk-v2`
- Give up dependency injection for `SesClient` and just build it manually. 

---

I think the best user experience would be achieved if we create individual modules per service. E.g. for [cloudfront service](https://console.aws.amazon.com/cloudfront). We will need a Gradle module named `aws-service-cloudfront`. 

For each module, the build file will be: 

```groovy
dependencies {
    api platform("software.amazon.awssdk:bom:$awssdkVersion")
    api 'software.amazon.awssdk:cloudfront'
    implementation project(":aws-sdk-v2")
    testImplementation 'software.amazon.awssdk:cloudfront'
}
```

Each module will have a single class. E.g.: 

```java
package io.micronaut.aws.sdk.v2.service.cloudfront;

import io.micronaut.aws.sdk.v2.service.AwsClientFactory;
import io.micronaut.context.annotation.Bean;
import io.micronaut.context.annotation.Factory;
import io.micronaut.context.annotation.Requires;
import java.lang.Override;
import javax.inject.Singleton;
import software.amazon.awssdk.auth.credentials.AwsCredentialsProviderChain;
import software.amazon.awssdk.http.SdkHttpClient;
import software.amazon.awssdk.http.async.SdkAsyncHttpClient;
import software.amazon.awssdk.regions.providers.AwsRegionProviderChain;
import software.amazon.awssdk.services.cloudfront.CloudFrontAsyncClient;
import software.amazon.awssdk.services.cloudfront.CloudFrontAsyncClientBuilder;
import software.amazon.awssdk.services.cloudfront.CloudFrontClient;
import software.amazon.awssdk.services.cloudfront.CloudFrontClientBuilder;

@Factory
public final class CloudFrontClientFactory extends AwsClientFactory<CloudFrontClientBuilder, CloudFrontAsyncClientBuilder, CloudFrontClient, CloudFrontAsyncClient> {
  protected CloudFrontClientFactory(AwsCredentialsProviderChain credentialsProvider,
      AwsRegionProviderChain regionProvider) {
    super(credentialsProvider, regionProvider);
  }

  @Override
  protected CloudFrontClientBuilder createSyncBuilder() {
    return CloudFrontClient.builder();
  }

  @Override
  @Singleton
  public CloudFrontClientBuilder syncBuilder(SdkHttpClient httpClient) {
    return super.syncBuilder(httpClient);
  }

  @Override
  @Singleton
  @Bean(
      preDestroy = "close"
  )
  public CloudFrontClient syncClient(CloudFrontClientBuilder builder) {
    return super.syncClient(builder);
  }

  @Override
  protected CloudFrontAsyncClientBuilder createAsyncBuilder() {
    return CloudFrontAsyncClient.builder();
  }

  @Override
  @Singleton
  @Requires(
      beans = SdkAsyncHttpClient.class
  )
  public CloudFrontAsyncClientBuilder asyncBuilder(SdkAsyncHttpClient httpClient) {
    return super.asyncBuilder(httpClient);
  }

  @Override
  @Singleton
  @Requires(
      beans = SdkAsyncHttpClient.class
  )
  @Bean(
      preDestroy = "close"
  )
  public CloudFrontAsyncClient asyncClient(CloudFrontAsyncClientBuilder builder) {
    return super.asyncClient(builder);
  }
}

```

 a test: 

```groovy
package io.micronaut.aws.sdk.v2.service.cloudfront

import io.micronaut.context.ApplicationContext
import software.amazon.awssdk.services.cloudfront.CloudFrontAsyncClient
import software.amazon.awssdk.services.cloudfront.CloudFrontClient
import spock.lang.AutoCleanup
import spock.lang.Shared
import spock.lang.Specification

class CloudFrontClientSpec extends Specification {

    @AutoCleanup
    @Shared
    ApplicationContext applicationContext = ApplicationContext.run()

    void "it can configure a sync client"() {
        when:
        CloudFrontClient client = applicationContext.getBean(CloudFrontClient)

        then:
        client.serviceName() == CloudFrontClient.SERVICE_NAME
    }

    void "it can configure an async client"() {
        when:
        CloudFrontAsyncClient client = applicationContext.getBean(CloudFrontAsyncClient)

        then:
        client.serviceName() == CloudFrontClient.SERVICE_NAME
    }
}
```

and an entry in the docs:

```
To use a SES client, add the following dependency:

dependency:ses[groupId="software.amazon.awssdk"]

Then, the following beans will be created:

* `software.amazon.awssdk.services.cloudfront.CloudFrontClient`
* ` software.amazon.awssdk.services.cloudfront.CloudFrontClientBuilder`

...

The HTTP client, credentials and region will be configured as per described in the <<sdkv2, SDK v2 documentation>>.
```

The user will just have to add a single dependency to use it: 

`implementation("io.micronaut.aws:micronaut-aws-service-ses")`

Alternative we could build the factories leveraging `@SdkClient` and the     
`annotationProcessor project(":aws-sdk-v2-processor")` @pgressa 

Not sure if the complexity of the annotation processor is worth it. 

```java
package io.micronaut.aws.sdk.v2.service.cloudfront;

import io.micronaut.aws.sdk.v2.service.SdkClient;
import io.micronaut.core.annotation.Internal;
import software.amazon.awssdk.services.cloudfront.CloudFrontAsyncClient;
import software.amazon.awssdk.services.cloudfront.CloudFrontAsyncClientBuilder;
import software.amazon.awssdk.services.cloudfront.CloudFrontClient;
import software.amazon.awssdk.services.cloudfront.CloudFrontClientBuilder;

import javax.inject.Singleton;

@SdkClient(client = CloudFrontClient.class, clientBuilder = CloudFrontClientBuilder.class, asyncClient = CloudFrontAsyncClient.class, asyncClientBuilder = CloudFrontAsyncClientBuilder.class)
@Singleton
@Internal
final class SdkClientCloudFront {
}
```

Maybe we don't need to create the 256 modules. Just create the top ones and document how to create a Factory in the users application for a module we don't have.

graemerocher:
What about placing all the clients in another module ie `aws-sdk-v2-clients` if folks want to take one the whole thing they can if not they are not forced to?

ilopmar:
I agree with @sdelamo here. I think that putting all the clients in one module (even if it's optional) can lead to performance issues in AWS Lambda because the environment needs to load a bunch of classes that won't be needed and the users don't really need if they only want to use SES (or any other single service).

pgressa:
Yeah, as said, it's lot of beans, though I think we could somehow help users to create those clients easily. 

0] to keep it there the way like now it's definitely out of question

1] generate module per client llike @sdelamo -> I think the problems will out-weight the value here. 257 more modules  to release and to create the documentation (that single page would be unreadable).. to have just a factory bean there.

2] I would prefer using `@SdkClient` annotation only here. We can address the complexity since not every time the user needs both sync/async. Also the BuilderClass could be derived from the client class, so we could have `@SdkClient` and `@SdkAsyncClient` with only a client class as a parameter field.

Plus, I believe the usage of the clients could be definitely ordered, so we could provide like top 10 clients out of the box in core. Finally with more AWS related modules, we could add more clients in specific modules like for example aws-param-store would contain the ssm plugin.

3] To have separated module like @graemerocher suggest -> I'm still concerned by the performance impact. Like it's still 256 class not found exceptions for single s3 bean.

sdelamo:
My proposal is:

Have modules for:

- The most common services 
- Services for which we do something else apart the factory (E.g. modules which do anything around distributed configuration).  

For example:

- `micronaut-aws-service-s3`
- `microanut-aws-service-dynamodb`
- `micronaut-aws-service-ses`
- `micronuat-aws-service-ssm`
- `micronaut-aws-service-secret-manager`

Those modules will have a transitive dependency to `aws-sdk-v2` where the common interface for the factory is defined.  `aws-sdk-v2` will be an extremely lightweight dependency. 

Then, in the docs, add an example of how to create a factory in your project for a service for which we don't provide module 

For this simple solution, I advocate not use `@SdkClient`, I prefer:

- having 6 simple factories implementing a common interface interface
- Than 6 classes with a @SdkClient annotation and the `aws-sdk-v2-processor` module which adds complexity with Java poet. 

Advantages: 

- For a user point of view, they will just need to add one dependency. If they want to use SES then they add  `micronaut-aws-service-ses` and they can inject a bean of type `SesClient`. The dependency which they add to their projects is lightweight. 

For us, it is easy to maintain since we are just releasing six modules, not distributing 256 modules or a FAT module with every service. 

Disadvantages: 

For less common services, the users will have to create a factory in their project. However, the steps to do that will be document and it involves writing a class annotated with `@Factory` which implements an interface.

pgressa:
Based on team discussion we agreed:
 - we will separate the existing client factory beans into standalone modules
 - the `SdkClient` annotation will be used as an alternative way how to easily client bean factory. (`SdkClient` will be both class and field annotation).

Issue #905, opened 2021-02-01
Add support for AWS Cloud Map service discovery


alvarosanchez:
[AWS Cloud Map](https://aws.amazon.com/blogs/aws/aws-cloud-map-easily-create-and-maintain-custom-maps-of-your-applications/) is a new service discovery service announced at re:Invent 2018.

Docs: https://docs.aws.amazon.com/cloud-map/index.html

musketyr:
Hi, we recently switched to using CloudMap for a service discovery and I just want to make sure the there is no support yet despite the merged PR above.

alvarosanchez:
Correct, Cloud Map isn't supported yet.

musketyr:
What is required for the implementation? Maybe we can work on this together?

graemerocher:
an implementation of `DiscoveryClient` similar to https://github.com/micronaut-projects/micronaut-aws/pull/1068/files#diff-86908c84f5c8c18116ad34b27c57a7607025fefe77a68b94a2b9dd4d86a309e9R58 that calls the cloud map sdk I imagine

musketyr:
thank you @graemerocher!

PR #1068, opened 2021-05-19
Service discovery with SDK v2

dstepanov:


dstepanov:
I left the previous route53 implementation untouched, it can be deleted if this is merged and rote53 us removed from parameter store dependency.

graemerocher:
@sdelamo can you take a look

aws-service-discovery/src/main/java/io/micronaut/discovery/aws/servicediscovery/registration/AwsServiceDiscoveryRegistrationClient.java
```diff
@@ -0,0 +1,290 @@
...
+            // try the metadata
+            if (instance.getMetadata().contains("instanceId")) {
+                opt = Optional.of(instance.getMetadata().asMap().get("instanceId"));
+            } else {
+                LOG.error("Cannot determine the instance ID. Are you sure you are running on AWS EC2?");
+            }
+        }
+
+        opt.ifPresent(instanceId -> {
+            if (discoveryService != null && discoveryService.healthCheckConfig() != null) {
```
dstepanov, line 155: Any idea what is `discoveryService` supposed to be? It's only set in tests but it looks like it should be provided from somewhere.

graemerocher: good question

dstepanov: I did more digging and it used to be set during the registration:
```
discoveryService = getService(route53AutoRegistrationConfiguration.getAwsServiceId());
```
it was removed in the initial version of this class.

sdelamo (changes requested):
@dstepanov could you add documentation ?

sdelamo:
Is this related to https://github.com/micronaut-projects/micronaut-aws/issues/905 ?

dstepanov:
My goal is to migrate to SDK v2. I think I chose the correctly API

aws-service-discovery/src/main/java/io/micronaut/discovery/aws/servicediscovery/AwsServiceDiscoveryClientConfiguration.java
```diff
@@ -0,0 +1,122 @@
...
+     *
+     * @param awsServiceId service id
+     */
+    public void setAwsServiceId(String awsServiceId) {
+        this.awsServiceId = awsServiceId;
+    }
+
+    @Override
+    public ConnectionPoolConfiguration getConnectionPoolConfiguration() {
+        return null;
```
graemerocher, line 120: Is this supposed to return `null`?

graemerocher (approved):
Looks good, just one query

graemerocher:
Ah yeah and documentation

dstepanov:
@sdelamo Where is the existing documentation for the service discovery module?

aws-service-discovery/src/main/java/io/micronaut/discovery/aws/servicediscovery/AwsServiceDiscoveryClientConfiguration.java
dstepanov: It's a copy of `Route53ClientDiscoveryConfiguration` and it looks like pooling is not supported for the client.

Issue #1164, opened 2021-08-24
String POST parameters get populated with full body


rnielsen:
### Expected Behavior

A POST method like

```
@Post("/stringParams")
public Map stringParams(String a, String b) {
    return Map.of("a", a, "b", b);
}
```
when invoked with a body of `{"a":"A","b":"B","c":"C"}` should populate parameters a and b with A and B respectively.

### Actual Behaviour

a and b are both being populated with the full JSON body. eg  `{"a":"A","b":"B","c":"C"}` 

Changing build.gradle to runtime('netty') instead of runtime('lambda') produces the expected result.

### Steps To Reproduce

Run `gradle test` on included project or `gradle run` and use curl
```
curl http://localhost:8080/stringParams -H 'Content-Type:application/json' --data '{"a":"A", "b":"B", "c":"C"}'
```

### Environment Information

- MacOS
- Java 11

### Example Application

https://github.com/rnielsen/micronaut-param

### Version

3.0.0

sdelamo:
I always create POJOs to encapsulate the body payloads. 

@rnielsen this is a workaround  which will work for you:

```groovy
Map stringParams(@Body BodyPogo bodyPogo) 
```

```groovy
package com.example

import io.micronaut.core.annotation.Introspected

@Introspected
class BodyPogo {
    String a
    String b
}

```

PR #1165, opened 2021-08-24
fix: Move body decoding above arguments

rnielsen:
Fix for #1164. Parameters were being populated with full JSON body rather than by name because the body was being decoded after populating the arguments.


function-aws-api-proxy/src/test/groovy/io/micronaut/function/aws/proxy/BodyArgumentsSpec.groovy
```diff
@@ -0,0 +1,49 @@
...
+        response.statusCode == 200
+        response.body == 'A'
+    }
+
+    @Controller('/body-arguments-test')
+    @Requires(property = 'spec.name', value = 'BodyArgumentsSpec')
+    static class BodyController {
+
+        @Post(uri = "/getA")
+        String getA(String a) {
```
sdelamo, line 45: Could you annotate the method with  `@Produces(MediaType.TEXT_PLAIN)` since the controller method is not returning JSON.

sdelamo (changes requested):
Thanks for the contribution. Could you implement the suggestion and I will get the PR merged. Thanks!

function-aws-api-proxy/src/test/groovy/io/micronaut/function/aws/proxy/BodyArgumentsSpec.groovy
sdelamo: It is unrelated to your PR. it is Kotlin and JDK 16.

PR #1175, opened 2021-09-08
fix: `function` `lamdba` default environments

### LINKED ISSUE #1171: Disabling environment deduction does not work with Lambda
### Expected Behavior

I expected environment to be just `ec2` 

### Actual Behaviour

environments are `ec2` `cloud` `function` `lambda`

### Steps To Reproduce

Create a Lambda function with the following environment variables: 

- `MICRONAUT_ENVIRONMENTS` to `ec2`
- `MICRONAUT_ENV_DEDUCTION` to `false`

### Environment Information

_No response_

### Example Application

_No response_

### Version

3.0.0

sdelamo:
Closes #1171

Currently `lambda` and `function` are set and the user cannot disable those via system property or environment property (the latter is probably what they want to do as that it is easy to do in the AWS Lambda console). When you deploy to Lambda `lambda` `function` `ec2` and `cloud` are the active environments.

This PR considers  `lambda` and `function` default environments. Hence, if the user specifies for example `MICRONAUT_ENVIRONMENTS=prod` in the AWS console, the active environment for the lambda function will be only `prod`.

The only drawback is that right now in the test classpath `function`, `lambda` and `test` are the active environments with PR only `test` is the active enviroments. Default enviroments are not used if an environment is deduced.

I consider the current behaviour a bug and I think we should fix it even if we remove `lambda` and `function` for the tests.

### DIFF
--- a/function-aws-alexa/src/main/java/io/micronaut/function/aws/alexa/AlexaFunction.java
+++ b/function-aws-alexa/src/main/java/io/micronaut/function/aws/alexa/AlexaFunction.java
@@ -67,7 +67,8 @@ public class AlexaFunction implements RequestStreamHandler, AutoCloseable, Close
     @SuppressWarnings("unchecked")
     @NonNull
     protected ApplicationContextBuilder newApplicationContextBuilder() {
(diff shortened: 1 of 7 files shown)


sdelamo:
We discussed in the engineering meeting and this could be consider a breaking change. `lambda` environment is added automatically because some optimisations are done if that environment are present.

PR #1188, opened 2021-09-28
Don't close application context between requests in MicronautRequestStreamHandler

### LINKED ISSUE #1187: MicronautStreamHandler closes MongoClient after processing request which is undesirable
### Discussed in https://github.com/micronaut-projects/micronaut-aws/discussions/1177

<div type='discussions-op-text'>

<sup>Originally posted by **mayurtanna** September 11, 2021</sup>
Hi @sdelamo 

Like I mentioned in the issue https://github.com/micronaut-projects/micronaut-aws/issues/1077, when I use MicronautRequestStreamHandler, it seems to be destroying the MongoClient (so ultimately connectionpool) once processing the request and hence we get an exception like below for subsequent request.

So, what I believe is it might be due to StreamFunctionExecutor class's execute method which is getting called internally when we use MicronautRequestStreamHandler. In that method it is closing the applicationContext in finally clause which seem to be destroying MongoClient internally.

I would like to understand, is it desirable to close the applicationContext post handling the request ? Because in case of MicronautRequestHandler I didn't find like we are closing applicationContext post handling of the request. Also, what should I do to keep the connection open ? 

Can you provide your thoughts here ?

----------------Exception---------------
Connection pool shut down: java.lang.IllegalStateException
java.lang.IllegalStateException: Connection pool shut down
at org.apache.http.util.Asserts.check(Asserts.java:34)
at org.apache.http.impl.conn.PoolingHttpClientConnectionManager.requestConnection(PoolingHttpClientConnectionManager.java:269)
at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
at java.base/java.lang.reflect.Method.invoke(Unknown Source)
at software.amazon.awssdk.http.apache.internal.conn.ClientConnectionManagerFactory$Handler.invoke(ClientConnectionManagerFactory.java:8</div>

sdelamo:
Close #1187

### DIFF
--- a/function-aws/build.gradle
+++ b/function-aws/build.gradle
@@ -5,4 +5,8 @@ dependencies {
     api "com.amazonaws:aws-lambda-java-core:1.2.1"
 
     testImplementation "io.micronaut:micronaut-inject-groovy"
+    testImplementation("io.micronaut.mongodb:micronaut-mongo-sync")
+    testImplementation("org.testcontainers:spock")
+    testImplementation("org.testcontainers:mongodb")
+    testImplementation("org.testcontainers:testcontainers")
 }
--- a/function-aws/src/main/java/io/micronaut/function/aws/MicronautRequestStreamHandler.java
+++ b/function-aws/src/main/java/io/micronaut/function/aws/MicronautRequestStreamHandler.java
@@ -85,4 +85,10 @@ public class MicronautRequestStreamHandler extends StreamFunctionExecutor<Contex
         String functionName = super.resolveFunctionName(env);
         return (functionName != null) ? functionName : ctxFunctionName;
     }
+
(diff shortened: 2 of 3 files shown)


sdelamo:
@graemerocher I added a test which verifies the scenario described at #1187 

I think that the fix is as easy as not closing the application context in `MicronautRequestStreamHandler::close`

Can you check it out?

ilopmar (approved):

graemerocher (approved):

PR #1194, opened 2021-10-05
fix: don't apply security filter again

### LINKED ISSUE #1166: After upgrading to Micronaut 3, Internal Server Error when request is not authorized
### Expected Behavior

Micronaut application with `micronaut-security`, when running on Lambda runtime, should respond with 401 or 403 when the authentication of the request fails.

### Actual Behaviour

After upgrading to Micronaut 3, it responds with 500.

### Steps To Reproduce

(See [example app](https://github.com/vmj/mn-aws-issue-1166) where steps 1-3 have already been taken, and you only need to run the test from step 4.)

1. Follow the guide https://guides.micronaut.io/latest/mn-application-aws-lambda-java11-gradle-groovy.html
2. After completing the guide, add the following dependencies for `micronaut-security`:
```
    compileOnly("io.micronaut.security:micronaut-security-annotations")
    implementation("io.micronaut.security:micronaut-security-jwt")
```
3. And add the `@Secured(SecurityRule.IS_AUTHENTICATED)` on top of the `BookController`
4. Now
   * either run the `BookControllerSpec` and expect it to fail because response status is 401 or 403 instead of 200
   * or deploy the app and use the AWS Lambda console to test invoke the function (using the same, unauthorized, test template as in the guide)

The test fails because the response code is 500.  The log shows:

```
ERROR i.m.f.a.p.AbstractLambdaContainerHandler - Error while handling request
io.micronaut.security.authentication.AuthorizationException: null
	at io.micronaut.security.filters.SecurityFilter.lambda$checkRules$7(SecurityFilter.java:175)
...
```

### Environment Information

- Operating System: MacOSX (for the spock test), AWS Lambda (when deployed)
- JDK Version: 11 (corretto)

### Example Application

https://github.com/vmj/mn-aws-issue-1166

### Version

3.0.1

sdelamo:
Close #1166 

I don't think this is the right fix. However, it fullfil the test. 

I tried not applying the filters after an exception but that breaks all the CORS headers valid present even if exception.

### DIFF
--- a/function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautLambdaContainerHandler.java
+++ b/function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautLambdaContainerHandler.java
@@ -49,6 +49,7 @@ import io.micronaut.http.server.exceptions.ExceptionHandler;
 import io.micronaut.http.server.types.files.StreamedFile;
 import io.micronaut.inject.qualifiers.Qualifiers;
 import io.micronaut.jackson.codec.JsonMediaTypeCodec;
+import io.micronaut.security.filters.SecurityFilter;
 import io.micronaut.web.router.*;
 import io.micronaut.web.router.resource.StaticResourceResolver;
 import org.apache.commons.io.IOUtils;
@@ -69,6 +70,7 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.stream.Stream;
 
 /**
@@ -556,8 +558,12 @@ public final class MicronautLambdaContainerHandler
             AtomicReference<HttpRequest<?>> requestReference,
             Publisher<? extends MutableHttpResponse<?>> routePublisher) {
         Publisher<? extends io.micronaut.http.MutableHttpResponse<?>> finalPublisher;
-        List<HttpFilter> filters = new ArrayList<>(lambdaContainerEnvironment.getRouter().findFilters(requestReference.get()));
+        HttpRequest<?> request = requestReference.get();
+        List<HttpFilter> filters = new ArrayList<>(lambdaContainerEnvironment.getRouter().findFilters(request));
         if (!filters.isEmpty()) {
+            if (request.getAttribute(SecurityFilter.KEY).isPresent()) {
+                filters.removeIf(httpFilter -> httpFilter instanceof SecurityFilter);
+            }
             // make the action executor the last filter in the chain
             filters.add((HttpServerFilter) (req, chain) -> (Publisher<MutableHttpResponse<?>>) routePublisher);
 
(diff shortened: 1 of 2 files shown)


sdelamo:
I talked yesterday with @jameskleeh about this. The solution is to use `RouteExecutor` from core. Starting from 3.1 is a singleton. It is used in `RoutingInBoundHandler`. We should attempt to copy this test file [FilterErrorSpec](https://github.com/micronaut-projects/micronaut-core/blob/3.0.x/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/filters/FilterErrorSpec.groovy) to this project.

vmj:
Thank you for looking into this @sdelamo .  This issue prevents us from upgrading from 2.x.

> The solution is to use `RouteExecutor` from core. Starting from 3.1 is a singleton. It is used in `RoutingInBoundHandler`. We should attempt to copy this test file [FilterErrorSpec](https://github.com/micronaut-projects/micronaut-core/blob/3.0.x/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/filters/FilterErrorSpec.groovy) to this project.

I take it you're pondering here about the "real fix" instead of the change in this PR.  But if this was a suggestion for a workaround to be used in an application, I didn't understand the instructions.

sdelamo:
Yes, we are working on the proper fix. 

 @mattmoss is working on this. @vmj will update you once a fix is ready and a new version is released. Thanks for your patience.

function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautLambdaContainerHandler.java
```diff
@@ -272,162 +306,118 @@ protected void handleRequest(
...
-                                containerResponse.close();
-                                return;
-                            }
-                        }
+    private MicronautAwsProxyResponse<?> toAwsProxyResponse(
+            MicronautAwsProxyResponse<?> response,
+            HttpResponse<?> message) {
 
-                        final Optional<URL> staticMatch = resourceResolver.resolve(containerRequest.getPath());
-                        if (staticMatch.isPresent()) {
```
vyusuf, line 367: I think this change has introduced a regression which stops the swagger integration from working since the documentation states it uses static resources.

i.e.
```
micronaut:
    router:
        static-resources:
            swagger:
                paths: classpath:META-INF/swagger
                mapping: /swagger/**
```

If there is an alternative for how this is meant to work, then the documentation needs to be updated.

sdelamo: can you create an issue which steps to reproduce?

lancekf: Hi, I'm running into this issue with generic static resources, not particularly swagger. I'm using micronaut-parent 3.7.4 and when I run the app locally, static resources work fine. When I deploy the app to AWS Lambda, I get a 404 for the same request(s). Why was this code removed?

driverpt: Is there any ETA on this issue?

PR #1206, opened 2021-11-02
Support custom key retrival and getting custom Versions of secrets

MatejNedic:
Allow users to name which keys to fetch instead of loading it by application name. This would provide bigger versatility and would allow users to use common credentials for POC or something like that. Not only that but this would allow users to fetch secrets by namings they choose and wouldn't have to switch to application name namings.
This integration would allow users to fetch versions of older parameters or secrets which could be something they might choose.

example of use:
```
application.properties
'aws.distributed-configuration.secrets': 'SECRETS:path/mongoDBPassword;path/kafkaSSLKey:version'
```
If needed to keep the old way of loading  we could just place an empty string after `SECRETS:`
```
application.properties
'aws.distributed-configuration.secrets': 'SECRETS:'
```

TO:DO
(If the idea is accepted would be happy to do all under)
Introduce loading by env which is currently not supported. (Each application-env file would be taken into account).
Introduce `aws.distributed-configuration` as a list and have a prefix change depending on the integration. (Parameter store could be easily integrated by just implementing one class).
Allow Optional before  SECRETS: meaning all secrets loaded will not cause failure on load.
If chosen old way of loading can be kept for backwards compatibility.
Clean code and namings more.


MatejNedic:
Hey @sdelamo , 
before going through with the implementation of to:dos would like to know your opinion on this approach for integration. Is this something that the Micronaut team would like to have or current way of loading is to stay? Would like to start working on PR if this is something you think is worth having. I understand it is breaking change and is whole another way of loading both parameter store and secrets manager.

PR #1208, opened 2021-11-11
build: update to use new plugins

sdelamo:


aws-bom/build.gradle
```diff
@@ -1,3 +1,10 @@
+plugins {
```
sdelamo, line 1: Are these the correct plugins for a `bom` submodule?

melix: For now this is correct, I didn't get to write a BOM plugin now, like for the module plugin.

melix (approved):
Nice, @sdelamo , I didn't expect anyone to start migrating now since this is work in progress and there's still a lot to untangle in the micronaut-build module but this is good to start and figure out where the problems are!

PR #1209, opened 2021-11-11
Send a user-agent header with requests to the Lambda Runtime API.

msailes:
An update to the AbstractMicronautLambdaRuntime and AwsLambdaRuntimeApi to send the version of Java and the version of Micronaut as the user-agent information in the HTTP request header.


graemerocher:
Thanks for the contribution!

PR #1216, opened 2021-11-13
Fix bom

sdelamo:
@melix 

> @sdelamo Just noticed that the AWS BOM is wrongly generated (I was using this project to try an internal BOM plugin). It contains projects it shouldn't, some groups are wrong and some versions undefined. It's late here so I won't fix it today but we have to fix it before a release.

bom was introduce here: 

https://github.com/micronaut-projects/micronaut-aws/pull/963

### DIFF
--- a/aws-bom/build.gradle
+++ b/aws-bom/build.gradle
@@ -4,10 +4,12 @@ javaPlatform {
 dependencies {
     constraints {
         for (Project p : rootProject.subprojects) {
+            if (p != project) {
+                evaluationDependsOn(p.path)
+            }
             if (p.name.endsWith("-bom")) {
                 continue
             }
-
             if (p.name.startsWith("test-suite")) {
                 continue
             }
--- a/aws-bom/build.gradle
+++ b/aws-bom/build.gradle
@@ -8,6 +8,9 @@ dependencies {
                 continue
             }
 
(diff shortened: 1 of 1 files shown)


sdelamo:
[I've excluded the `test-suite` modules in the 2.6.x](https://github.com/micronaut-projects/micronaut-aws/pull/1217). test-suite modules appeared first in 2.6.x branch. I have merged up from `2.6.x` to `2.7.x`, `2.8.x`, `2.9.x` `3.0.x` and `master`

Should the following snippet had been present since we introduced the bom module: 

```groovy
if (p != project) {
      evaluationDependsOn(p.path)
}
```

melix:
The `evaluationDependsOn` is needed because the BOM module directly accesses the other project's state (via `project.version`) which is unfortunately unsafe. Therefore it is possible that the `version` (or `group`) written in the BOM is `undefined` without it.

melix (approved):

sdelamo:
@melix  this should have been applied since we introduced the `bom` submodule right?

melix:
It is required as soon as you access another project's state directly. However, because of how Gradle evaluates `allprojects` and `subprojects`, it _may_ be that for builds which use this pattern, it works correctly. This is however undefined behavior, since there's no guarantee _when_ the closure will be called.

PR #1228, opened 2021-12-02
Upgrade to Micronaut build plugins 5.0.x

melix:


buildSrc/src/main/groovy/io.micronaut.build.internal.aws-test-suite.gradle
```diff
@@ -0,0 +1,11 @@
+/**
+ * Plugin for test suite projects.
+ */
+plugins {
+    id 'io.micronaut.build.internal.aws-base'
```
sdelamo, line 5: why apply the ` aws-base` plugin?. We don't need to specify a group or a version for the `test-suite` projects.

melix: Because it's cleaner from my POV if all modules have a group/version set, rather than the undefined Gradle default.

sdelamo: I don't know how to feel about this. Since it is not something we want to release I am intentionally leaving group and version blank for things we don't want release in the builds I am upgrading.

melix: The `maven-publish` plugin is not applied, so it won't be published in any case. Coordinates and publications are orthogonal. Basically this just gives an identifier to the project.

sdelamo: in general, your opinion is that any java / groovy / kotlin module should have a `group` and `version` independently if we want to publish it?

melix: Yes. Otherwise from an external point of view it looks like a mistake. This also makes it possible to perform dependency substitutions, although for a test suite it might be a bit weird use case :)

sdelamo (approved):

PR #1231, opened 2021-12-03
build: Gradle build plugin to 5.0.2 and set projectGroup

ilopmar:
This upgrades to 5.0.2 that fixes the snapshot publish

### DIFF
--- a/aws-alexa-httpserver/build.gradle
+++ b/aws-alexa-httpserver/build.gradle
@@ -1,5 +1,5 @@
 plugins {
-    id "io.micronaut.build.internal.aws-module"
+    id("io.micronaut.build.internal.module")
 }
 
 dependencies {
--- a/aws-alexa/build.gradle
+++ b/aws-alexa/build.gradle
@@ -1,5 +1,5 @@
 plugins {
-    id "io.micronaut.build.internal.aws-module"
+    id("io.micronaut.build.internal.module")
 }
 
 dependencies {
--- a/aws-bom/build.gradle
+++ b/aws-bom/build.gradle
@@ -1,4 +1,3 @@
 plugins {
-    id 'io.micronaut.build.internal.aws-base'
     id 'io.micronaut.build.internal.bom'
 }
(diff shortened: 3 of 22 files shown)


melix (approved):

ilopmar:
I'll leave @sdelamo the final approve + merge so he is aware of the changes

sdelamo (changes requested):
If I understand correctly, with 5.0.2 now version uses `projectVersion` in gradle.properties and `group` uses `projectGroup` in `gradle.properties`. 

If that it is correct, then I think modules should apply `io.micronaut.build.internal.module` not `io.micronaut.build.internal.aws-module.gradle`

melix:
> If that it is correct, then I think modules should apply `io.micronaut.build.internal.module` not io.micronaut.build.internal.aws-module.gradle

There _can_ apply only the `io.micronaut.build.internal.module` plugin but there's no harm in keeping the `aws-module` plugin in case some common configuration shows up (this is what the project template actually says now: it has a convention plugin but explains you can remove it if there's no common configuration).

sdelamo:
> > If that it is correct, then I think modules should apply `io.micronaut.build.internal.module` not io.micronaut.build.internal.aws-module.gradle
> 
> There _can_ apply only the `io.micronaut.build.internal.module` plugin but there's no harm in keeping the `aws-module` plugin in case some common configuration shows up (this is what the project template actually says now: it has a convention plugin but explains you can remove it if there's no common configuration).

In this PR, 

https://github.com/micronaut-projects/micronaut-problem-json/pull/80/files

We delete `problem-base`. 

I strongly believe that we have to do the same strategy in every module. 

I personally prefer to delete the base plugin if the only thing that it does is to apply `internal.module`.

melix:
So objection from my side.

sdelamo:
Why did you close the PR?. I still think we should use `projectGroup=io.micronaut.aws`and upgrade to 5.0.2

melix:
Closing wasn't intentional at all. I don't even recall clicking a button :/

Issue #1287, opened 2022-02-23
MicronautLambdaHandler should populate diagnostic aws values for slf4j


theaspect:
### Feature description

`MicronautRequestHandler` has `populateMappingDiagnosticContextValues` but `MicronautLambdaHandler` for some reason does not.

If there're no objection I could prepare PR replicating the same method from RequestHandler. Or probably it make sense to abstract it into general class.

graemerocher:
PRs welcome

PR #1310, opened 2022-03-11
Register Lambda Context as singletons and MDC for every Handler

theaspect:
Fixes: https://github.com/micronaut-projects/micronaut-aws/issues/1332
Fixes: https://github.com/micronaut-projects/micronaut-aws/issues/1326

This is first attempt to generalize adding diagnostic information to lambda handler despite significant difference between `MicronautRequestHandler` and `MicronautLambdaHandler`. See also #1287 

I'll update documentation and add test when receive a green light for choosen approach.

@graemerocher and @sdelamo your comments are welcome especially regarding bean initialization and registration.

### DIFF
--- a/function-aws/src/main/java/io/micronaut/function/aws/DiagnosticInfoPopulator.java
+++ b/function-aws/src/main/java/io/micronaut/function/aws/DiagnosticInfoPopulator.java
@@ -1,0 +1,109 @@
+package io.micronaut.function.aws;
+
+import com.amazonaws.services.lambda.runtime.ClientContext;
+import com.amazonaws.services.lambda.runtime.CognitoIdentity;
+import com.amazonaws.services.lambda.runtime.LambdaLogger;
+import io.micronaut.context.ApplicationContext;
+import io.micronaut.core.annotation.Introspected;
+import io.micronaut.core.annotation.NonNull;
+import io.micronaut.core.util.StringUtils;
+import org.slf4j.MDC;
+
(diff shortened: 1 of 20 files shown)


function-aws/src/main/java/io/micronaut/function/aws/DiagnosticInfoPopulator.java
```diff
@@ -0,0 +1,109 @@
...
+import io.micronaut.context.ApplicationContext;
+import io.micronaut.core.annotation.Introspected;
+import io.micronaut.core.annotation.NonNull;
+import io.micronaut.core.util.StringUtils;
+import org.slf4j.MDC;
+
+import java.util.Optional;
+
+@Introspected
+public class DiagnosticInfoPopulator {
```
graemerocher, line 15: not sure why this is `@Introspected`? It doesn't have any properties

graemerocher:
General approach seems fine

sdelamo:
@theaspect thanks for the contribution. I will review it during this week and gets this merged.

sdelamo:
I have created two beans: 

`DiagnosticInfoPopulator`
`LambdaContextFactory` 

so that users can provide their own implementations if what we do is not what they are looking for. 

These PR tests that  Lambda `Context` beans are registered. 

It also tests how to inject the Lambda `Context`. I added a docs section.

graemerocher (approved):

PR #1323, opened 2022-03-29
Move away from Jackson

timyates:
Started with function-aws-api-proxy to gather feedback

- Is this going in the right direction?
- Are request bodies always Maps in Lambdas?  I couldn't find a way to [parse a JsonNode here](https://github.com/micronaut-projects/micronaut-aws/compare/update-build...remove-jackson?expand=1#diff-dd8cf116b1d2f8e327fcac6c6f62b5464b6d52d16c6a8563be369b7c6fdea1a1L428-R446)
- I had to @SerdeImport` quite a few AWS classes, but I'm not sure our coverage has shown them all
- I had to write my own replacements for [ObjectWriter and ObjectReader](https://github.com/micronaut-projects/micronaut-aws/compare/update-build...remove-jackson?expand=1#diff-dcd9cc3e7ce4d187698e4488748aae2c521f3702fe1f6d8224bb6997c909cf9bR260-R270)

When I look at a scan for running the tests, I still see a LOT of Jackson, so I'm not sure if I'm doing the right thing

https://ge.micronaut.io/s/s7dmxjatvn6c4/dependencies?dependencies=fasterxml&expandAll

### DIFF
--- a/function-aws-custom-runtime/build.gradle
+++ b/function-aws-custom-runtime/build.gradle
@@ -11,7 +11,6 @@ dependencies {
     api libs.micronaut.http.client
     api libs.managed.aws.lambda.events
     implementation libs.micronaut.serde
-    implementation libs.spotbugs.annotations
 
     testAnnotationProcessor libs.micronaut.inject.java
     testAnnotationProcessor libs.micronaut.serde.processor


function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/AbstractLambdaContainerHandler.java
```diff
@@ -119,7 +117,7 @@ protected AbstractLambdaContainerHandler(Class<RequestType> requestClass,
      *
      * @return Return the object mapper.
      */
-    protected abstract ObjectMapper objectMapper();
+    protected abstract JsonMapper objectMapper();
```
sdelamo, line 120: There is a `import io.micronaut.serde.ObjectMapper;`. Why not return it instead?

function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautLambdaContainerContext.java
```diff
@@ -40,9 +40,9 @@
     JsonMediaTypeCodec getJsonCodec();
 
     /**
-     * @return The Jackson's {@link ObjectMapper}
+     * @return The {@link JsonMapper}
      */
-    default ObjectMapper getObjectMapper() {
-        return getJsonCodec().getObjectMapper();
+    default JsonMapper getObjectMapper() {
```
sdelamo, line 45: There is a `import io.micronaut.serde.ObjectMapper;` why not use instead of `JsonMapper`?

function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautLambdaContainerHandler.java
```diff
@@ -202,18 +220,18 @@ public ApplicationContext getApplicationContext() {
     }
 
     @Override
-    protected ObjectMapper objectMapper() {
+    protected JsonMapper objectMapper() {
```
sdelamo, line 223: why not use `import io.micronaut.serde.ObjectMapper` instead?

sdelamo (changes requested):
This will be a breaking change. We have to do it for next mayor of AWS.

sdelamo:
@yawkat can you provide feedback?

function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautLambdaContainerContext.java
timyates: JsonCodec returns a JsonMapper, which isn't an ObjectMapper 😔

function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautLambdaContainerHandler.java
```diff
@@ -422,12 +444,10 @@ private void decodeRequestBody(MicronautAwsProxyRequest<?> containerRequest, Rou
...
                                 final Object decoded = lambdaContainerEnvironment.getJsonCodec().decode(bodyArgument, body.get());
-                                ((MicronautAwsProxyRequest) containerRequest)
-                                        .setDecodedBody(decoded);
+                                ((MicronautAwsProxyRequest) containerRequest).setDecodedBody(decoded);
                             } else {
-                                final JsonNode jsonNode = lambdaContainerEnvironment.getJsonCodec().decode(JsonNode.class, body.get());
-                                ((MicronautAwsProxyRequest) containerRequest)
-                                        .setDecodedBody(jsonNode);
+                                final Map<?, ?> node = lambdaContainerEnvironment.getJsonCodec().decode(Map.class, body.get());
+                                ((MicronautAwsProxyRequest) containerRequest).setDecodedBody(node);
```
timyates, line 450: We can't decode to a serde JsonNode here (it fails to create one), so I fell back to a Map.  This will break if the body is not a map (ie: a simple json string), but I'm not sure if Lambdas can receive simple strings

timyates:
The HateOas model doesn't seem to be coming through correctly...

```
{"message":"Internal Server Error","_links":{"self":[{"href":"https://null.execute-api.us-east-1.amazonaws.com/filter-error-spec-3","templated":false}]},"_embedded":{"errors":[{}]}}
```

Notice the errors are not serialized with any properties...

I've tried `@SerdeImport(JsonError.class)` but it has no effect on the result

The model should look like this:

image

timyates:
@alvarosanchez Any idea how we [migrate this to Serde serialization](https://github.com/micronaut-projects/micronaut-aws/blob/8c54dd859c25e46acf1e3dbfa8045772bb3830f7/function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/AwsObjectMapperFactory.java#L33)?  The tests are checking for Jackson features which I don't believe we support in Serde

timyates:
@alvarosanchez Thanks!  We're well into new major version territory already, so that should be fine 👍

timyates:
So here https://github.com/micronaut-projects/micronaut-aws/blob/dca8a6cfe7a2325aac742034881d61d6cec64b56/aws-alexa-httpserver/src/test/groovy/io/micronaut/aws/alexa/httpserver/controllers/SkillControllerPathSpec.groovy#L42-L45

We send an AWS `RequestEnvelope`([link](https://github.com/alexa/alexa-apis-for-java/blob/master/ask-sdk-model/src/com/amazon/ask/model/RequestEnvelope.java)), and expect an AWS `ResponseEnvelope` in return.

Serde fails to deserialize the request model as `No default constructor exists`... I'm going to look into a custom deserializer

timyates:
@graemerocher @yawkat I'm not sure how to use Serde to handle RequestEnvelope from AWS (see https://github.com/micronaut-projects/micronaut-aws/pull/1323#issuecomment-1083124887 above)

The [class iteself](https://github.com/alexa/alexa-apis-for-java/blob/master/ask-sdk-model/src/com/amazon/ask/model/RequestEnvelope.java) and a lot of it's member classes use

```
@JsonDeserialize(builder = RequestEnvelope.Builder.class)
```

Which we don't support in Serde 🤔. Anyone got any ideas about getting round this?

yawkat:
@timyates i would use a `JsonCreator` in a mixin

timyates:
@yawkat Like this?

```
public class RequestEnvelopeMixin {

    @JsonCreator
    public static RequestEnvelope factory(
            @JsonProperty("version") String version,
            @JsonProperty("session") Session session,
            @JsonProperty("context") Context context,
            @JsonProperty("request") Request request
    ) {
        return RequestEnvelope.builder()
                .withVersion(version)
                .withSession(session)
                .withContext(context)
                .withRequest(request)
                .build();
    }
}
```

I guess I need to mixin the whole tree of types before it will start working 😀

graemerocher:
A custom deserializer is probably the simplest to be honest

yawkat:
@graemerocher the classes only get bigger, see eg https://github.com/alexa/alexa-apis-for-java/blob/master/ask-sdk-model/src/com/amazon/ask/model/Context.java

graemerocher:
seems like we are going to have to add support for `builder=`

timyates:
Currently blocked by https://github.com/micronaut-projects/micronaut-serialization/issues/176

lesteenman:
The blocking issue seems closed; Any intentions to continue work on this?

Issue #1330, opened 2022-04-01
Lamba Proxy does not populate response headers correctly


sdelamo:
### Issue description

In an `Application` with the `netty` runtime the following test passes: 

```
package com.example

import io.micronaut.context.annotation.Property
import io.micronaut.context.annotation.Requires
import io.micronaut.http.HttpRequest
import io.micronaut.http.HttpResponse
import io.micronaut.http.HttpStatus
import io.micronaut.http.MediaType
import io.micronaut.http.annotation.Controller
import io.micronaut.http.annotation.Get
import io.micronaut.http.annotation.Status
import io.micronaut.http.client.HttpClient
import io.micronaut.http.client.annotation.Client
import io.micronaut.test.extensions.spock.annotation.MicronautTest
import jakarta.inject.Inject
import spock.lang.Specification

@Property(name = 'spec.name', value = 'ContentTypeSpec')
@MicronautTest
class ContentTypeSpec extends Specification {

    @Inject
    @Client("/")
    HttpClient httpClient

    void "verify controllers return json by default"() {
        given:
        HttpRequest<?> request = HttpRequest._GET_('/json/bydefault').accept(MediaType._APPLICATION\_JSON_)

        when:
        HttpResponse<String> response = httpClient.toBlocking().exchange(request, String)

        then:
        HttpStatus._OK_ == response.status()
        response.body.get() == '{"msg":"Hello world"}'
        response.headers
        "application/json" == response.headers.get("Content-Type")
    }

    @Controller('/json')
    @Requires(property = 'spec.name', value = 'ContentTypeSpec')
    static class BodyController {

        @Get("/bydefault")
        @Status(HttpStatus._OK_)
        Map<String, Object> index() {
            [msg: "Hello world"]
        }
    }
}
```

However, an `Application`  with `lambda` runtime the following tests fails: 

```
package com.example

import com.amazonaws.serverless.proxy.internal.testutils.AwsProxyRequestBuilder
import com.amazonaws.serverless.proxy.internal.testutils.MockLambdaContext
import com.amazonaws.serverless.proxy.model.AwsProxyRequest
import com.amazonaws.serverless.proxy.model.AwsProxyResponse
import com.amazonaws.services.lambda.runtime.Context
import io.micronaut.function.aws.proxy.MicronautLambdaHandler
import io.micronaut.http.HttpHeaders
import io.micronaut.http.HttpMethod
import io.micronaut.http.HttpStatus
[...]

PR #1331, opened 2022-04-01
add failing test

sdelamo:
Fixes: https://github.com/micronaut-projects/micronaut-aws/issues/1330


sdelamo:
@timyates this PR contains a failing test. See that it is annotated with `@PendingFeature`. Can you investigate why the content type header in the response is not correctly populated.

timyates:
We do get the header back if we annotate

https://github.com/micronaut-projects/micronaut-aws/blob/cc49ea192a383d74ffb565459b75f2044f9dc7e1/function-aws-api-proxy/src/test/groovy/io/micronaut/function/aws/proxy/ContentTypeSpec.groovy#L57-L61

With

```
        @Produces(MediaType.APPLICATION_JSON)
```

But again, it's not copied over as the comparator in `toAwsProxyResponse` is excluding it

PR #1334, opened 2022-04-02
doc: Corretto with two r and two tt

okay19:


sdelamo:
@okay19 Thanks for the contribution

PR #1337, opened 2022-04-04
fix: proxy-test property sources to lambda handler ctx

sdelamo:
Without this PR, you cannot use a `@Requires` annotation for an `@Property` - the typical `spec.name` pattern to avoid bean pollution. Because the api-proxy-test appliation context property sources are differnet than the `MicronautLambdaHandler` property sources.


sdelamo:
@graemerocher can you review this?

PR #1347, opened 2022-04-13
feat: adds AWS CDK module

sdelamo:


msailes:
This looks fantastic!

aws-cdk/build.gradle
```diff
@@ -0,0 +1,15 @@
...
+
+dependencies {
+    api libs.aws.cdk.lib
+    api("io.micronaut.starter:micronaut-starter-api:$micronautVersion")
+    testImplementation(project(":function-aws-api-proxy"))
+}
+
+micronautBuild {
+    binaryCompatibility {
+        enabled = false
```
alvarosanchez, line 13: Why disable it?

aws-cdk/build.gradle
```diff
@@ -0,0 +1,15 @@
+plugins {
+    id("io.micronaut.build.internal.module")
+}
+
+dependencies {
+    api libs.aws.cdk.lib
+    api("io.micronaut.starter:micronaut-starter-api:$micronautVersion")
```
alvarosanchez, line 7: `api` or `implementation`?

alvarosanchez (commented):
So are we planning to support other CDK resources in the future? Because to me, this looks like an `aws-cdk-lambda` module.

aws-cdk/build.gradle
sdelamo: It is `api` intentionally. So that uses can provide both enums `BuildTool` and `ApplicatonType`.

aws-cdk/build.gradle
sdelamo: Since `aws-cdk` is a new module, it fails unless you disable it. 

```
> Task :aws-cdk:japiCmp FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':aws-cdk:japiCmp'.
> Could not resolve all files for configuration ':aws-cdk:detachedConfiguration2'.
   > Could not find any matches for io.micronaut.aws:micronaut-aws-cdk:+ as no versions of io.micronaut.aws:micronaut-aws-cdk are available.
     Searched in the following locations:
       - https://repo.maven.apache.org/maven2/io/micronaut/aws/micronaut-aws-cdk/maven-metadata.xml
     Required by:
         project :aws-cdk

* Try:
```

sdelamo:
> So are we planning to support other CDK resources in the future? Because to me, this looks like an `aws-cdk-lambda` module.

I placed them at package `io/micronaut/aws/cdk/function` instead of `io/micronaut/aws/cdk` in case we provide other utility classes for CDK. 

I you think it would be best to rename the artifact id from `micronaut-aws-cdk` to `micronaut-aws-cdk-lambda` or `micronaut-aws-cdk-function` happy to do it.

alvarosanchez (approved):

PR #1420, opened 2022-08-02
Allow overriding the endpoint via configuration

alvarosanchez:
Overriding the S3 endpoint is required when using services such as Localstack, or when using an S3 client for any S3-compatible storage service.

Before this change, users would need to implement a `BeanCreatedEventListener` to programmatically set the endpoint. This PR allows setting the endpoint via configuration

### DIFF
--- a/aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/service/s3/S3ConfigurationProperties.java
+++ b/aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/service/s3/S3ConfigurationProperties.java
@@ -22,6 +22,8 @@ import software.amazon.awssdk.services.s3.S3Client;
 import software.amazon.awssdk.services.s3.S3Configuration;
 import software.amazon.awssdk.services.s3.S3Configuration.Builder;
 
+import java.net.URI;
+
 /**
  * Configuration properties for S3.
  *
@@ -34,10 +36,26 @@ public class S3ConfigurationProperties extends AWSConfiguration {
     @ConfigurationBuilder(prefixes = {""}, excludes = {"profileFile", "applyMutation"})
     private Builder builder = S3Configuration.builder();
 
+    private URI endpointOverride;
+
     /**
      * @return The builder
      */
     public Builder getBuilder() {
         return builder;
+    }
+
+    /**
+     * @return The endpoint with which the AWS SDK should communicate
+     */
+    public URI getEndpointOverride() {
+        return endpointOverride;
+    }
+
+    /**
+     * @param endpointOverride The endpoint with which the AWS SDK should communicate
+     */
+    public void setEndpointOverride(URI endpointOverride) {
+        this.endpointOverride = endpointOverride;
     }
 }
--- a/aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/service/s3/S3ConfigurationProperties.java
+++ b/aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/service/s3/S3ConfigurationProperties.java
@@ -57,6 +57,7 @@ public class S3ConfigurationProperties extends AWSConfiguration {
     }
 
     /**
+     * Provide a URI to override the endpoint with which the AWS SDK should communicate. Optional. Defaults to `null`.
      * @param endpointOverride The endpoint with which the AWS SDK should communicate
      * @since 3.6.2
      */
--- a/aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/service/s3/S3ConfigurationProperties.java
+++ b/aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/service/s3/S3ConfigurationProperties.java
@@ -47,6 +47,7 @@ public class S3ConfigurationProperties extends AWSConfiguration {
 
     /**
      * @return The endpoint with which the AWS SDK should communicate
+     * @since 3.6.2
      */
     public URI getEndpointOverride() {
         return endpointOverride;
@@ -54,6 +55,7 @@ public class S3ConfigurationProperties extends AWSConfiguration {
 
     /**
      * @param endpointOverride The endpoint with which the AWS SDK should communicate
+     * @since 3.6.2
      */
     public void setEndpointOverride(URI endpointOverride) {
         this.endpointOverride = endpointOverride;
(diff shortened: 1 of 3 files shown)


aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/service/s3/S3ConfigurationProperties.java
```diff
@@ -34,10 +36,26 @@ public class S3ConfigurationProperties extends AWSConfiguration {
...
+
     /**
      * @return The builder
      */
     public Builder getBuilder() {
         return builder;
     }
+
+    /**
+     * @return The endpoint with which the AWS SDK should communicate
```
timyates, line 49: `@since`?

timyates (approved):

alvarosanchez:
Using a listener should be a last resort.

One line of configuration is better than 50 lines of listener code.

sdelamo:
> Using a listener should be a last resort.
> 
> One line of configuration is better than 50 lines of listener code.

Should this apply only to S3? or should we do it for the other factories?  Can you show me an example of what would you configure as the value? a [LocalStack URL](https://localstack.cloud/)

aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/service/s3/S3ConfigurationProperties.java
```diff
@@ -34,10 +36,28 @@ public class S3ConfigurationProperties extends AWSConfiguration {
...
+     * @since 3.6.2
+     */
+    public URI getEndpointOverride() {
```
sdelamo, line 52: annotate return type with `@Nullable`

aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/service/s3/S3ConfigurationProperties.java
```diff
@@ -34,10 +36,28 @@ public class S3ConfigurationProperties extends AWSConfiguration {
     @ConfigurationBuilder(prefixes = {""}, excludes = {"profileFile", "applyMutation"})
     private Builder builder = S3Configuration.builder();
 
+    private URI endpointOverride;
```
sdelamo, line 39: annotate  with `@Nullable`

aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/service/s3/S3ConfigurationProperties.java
```diff
@@ -34,10 +36,28 @@ public class S3ConfigurationProperties extends AWSConfiguration {
...
+     */
+    public URI getEndpointOverride() {
+        return endpointOverride;
+    }
+
+    /**
+     * @param endpointOverride The endpoint with which the AWS SDK should communicate
+     * @since 3.6.2
+     */
+    public void setEndpointOverride(URI endpointOverride) {
```
sdelamo, line 60: annotate parameter type with `@Nullable`.

aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/service/s3/S3ConfigurationProperties.java
```diff
@@ -34,10 +36,28 @@ public class S3ConfigurationProperties extends AWSConfiguration {
...
+    /**
+     * @return The endpoint with which the AWS SDK should communicate
+     * @since 3.6.2
+     */
+    public URI getEndpointOverride() {
+        return endpointOverride;
+    }
+
+    /**
+     * @param endpointOverride The endpoint with which the AWS SDK should communicate
```
sdelamo, line 57: Improve the setter description. Say that default value is `null`. The setter description  is used for description here: https://micronaut-projects.github.io/micronaut-aws/latest/guide/configurationreference.html#io.micronaut.aws.sdk.v2.service.s3.S3ConfigurationProperties

sdelamo (changes requested):

alvarosanchez:
> Should this apply only to S3? or should we do it for the other factories? 

S3 is the only service for which we have a `@ConfigurationProperties`

> Can you show me an example of what would you configure as the value? a [LocalStack URL](https://localstack.cloud/)

```groovy
@MicronautTest
@Testcontainers
class AwsS3BucketSpec extends ObjectStorageSpecification implements TestPropertyProvider {

    @Shared
    @AutoCleanup
    public LocalStackContainer localstack = new LocalStackContainer(DockerImageName.parse("localstack/localstack"), false)
            .withServices(S3)

    @Override
    Map<String, String> getProperties() {
        return [
                "aws.accessKeyId": localstack.getAccessKey(),
                "aws.secretKey": localstack.getSecretKey(),
                "aws.region": localstack.getRegion(),
                "aws.s3.endpoint-override": localstack.getEndpointOverride(S3)

        ]
    }

    //...

}
```

aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/service/s3/S3ConfigurationProperties.java
alvarosanchez: The description is taken verbatim from [the own AWS SDK](https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkClientBuilder.java#L57).

Also, are 3 `@Nullable` annotations not enough to indicate that this is null?

alvarosanchez:
@sdelamo why did you add this to a new minor version of the module? Are you going to release it immediately?

sdelamo:
> @sdelamo why did you add this to a new minor version of the module? Are you going to release it immediately?

yes. I will release maybe today. I have to release a patch and then I can release a minor with the CDK minor upgrade and this PR.

aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/service/s3/S3ConfigurationProperties.java
sdelamo: Yes, they are if you open the class. However, if you are reading the docs you have no idea if it is optional unless we tell them. I added a commit. Now it looks like: 
Screenshot 2022-08-02 at 17 04 25

sdelamo (approved):

alvarosanchez:
@sdelamo if you want to see this in action: https://github.com/micronaut-projects/micronaut-object-storage/pull/65

PR #1422, opened 2022-08-03
if the return type is void just return null

asantare:
there are no tests for the function lambdas (only 2 are run but are not testing the invocation), if you can point me out how to add one please let me know


graemerocher:
can you do the commits with the same account as you created the PR with? Also a test would be nice. Thanks.

asantare:
hi Graeme, I added the test method in the Java Spec, as in Groovy Spec it complains that the function has no return value (which is, well, mathematically correct!)

graemerocher:
Thanks for the contribution!

PR #1433, opened 2022-08-09
New AfterExecutionEvent for AWS Lambda

musketyr:
This is an event we discussed in https://github.com/micronaut-projects/micronaut-core/issues/7858. This can be used as a hook to finish any work which needs to be done before the Lambda function hibernates.


function-aws/src/main/java/io/micronaut/function/aws/MicronautRequestHandler.java
```diff
@@ -125,7 +126,14 @@ public final O handleRequest(I input, Context context) {
         if (!inputType.isInstance(input)) {
             input = convertInput(input);
         }
-        return this.execute(input);
+        try {
+            O output = this.execute(input);
+            applicationContext.getEventPublisher(AfterExecutionEvent.class).publishEvent(AfterExecutionEvent.success(output));
```
graemerocher, line 131: Need to cache the event publisher lookup since it can be expensive in terms of adding to the execution time

function-aws/src/main/java/io/micronaut/function/aws/event/AfterExecutionEvent.java
```diff
@@ -0,0 +1,87 @@
...
+    public static AfterExecutionEvent success(@Nullable Object output) {
+        return new AfterExecutionEvent(output, null);
+    }
+
+    /**
+     * Creates a new {@link AfterExecutionEvent} with an exception been thrown.
+     * @param exception the exception which has been thrown during the execution
+     * @return a new {@link AfterExecutionEvent} with an exception been thrown.
+     */
+    public static AfterExecutionEvent failure(RuntimeException exception) {
```
graemerocher, line 49: Why `RuntimeException`? Just using `Exception` should be ok

graemerocher (changes requested):

function-aws/src/main/java/io/micronaut/function/aws/event/AfterExecutionEvent.java
musketyr: I initially had `Throwable` but as `execute` method is not throwing any checked exceptions the only `RuntimeException` can happen under normal circumstances. I'm easy to broaden it to `Exception` or `Throwable` if you think it's better.

function-aws/src/main/java/io/micronaut/function/aws/MicronautRequestHandler.java
musketyr: actually the lookup is always happening only once either on success or on exception (if I skip the unlikely option that the lookup/publish fails)

dstepanov: It will lookup the bean again when the lambda is reused.

function-aws/src/main/java/io/micronaut/function/aws/event/AfterExecutionEvent.java
musketyr: nvm, I have noticed that `MicronautRequestStreamHandler` is actually throwing `IOException`.

musketyr:
@graemerocher I have moved obtaining the publisher into constructors but I got a following error which I don't understand

```
io.micronaut.context.exceptions.NoSuchBeanException: No bean of type [io.micronaut.context.event.ApplicationEventPublisher<io.micronaut.function.aws.event.AfterExecutionEvent>] exists. Make sure the bean is not disabled by bean requirements (enable trace logging for 'io.micronaut.context.condition' to check) and if the bean is enabled then ensure the class is declared a bean and annotation processing is enabled (for Java and Kotlin the 'micronaut-inject-java' dependency should be configured as an annotation processor).
	at io.micronaut.context.DefaultBeanContext.resolveBeanRegistration(DefaultBeanContext.java:2784)
	at io.micronaut.context.DefaultBeanContext.getBean(DefaultBeanContext.java:1596)
	at io.micronaut.context.DefaultBeanContext.getBean(DefaultBeanContext.java:865)
	at io.micronaut.context.BeanLocator.getBean(BeanLocator.java:94)
	at io.micronaut.context.BeanContext.getEventPublisher(BeanContext.java:68)
	at io.micronaut.function.aws.MicronautRequestStreamHandler.<init>(MicronautRequestStreamHandler.java:54)
	at io.micronaut.function.aws.MicronautStreamHandlerLambdaContextSpec.Lambda Context beans are registered(MicronautStreamHandlerLambdaContextSpec.groovy:23)
```

Shouldn't the publisher be always present?

musketyr:
FYI the last failing check is from Sonarqube (maybe it's not configured for PRs?)

sdelamo (changes requested):
I have created a PR https://github.com/musketyr/micronaut-aws/pull/1. 

The PR misses documentation. We should document the existence of `AfterExecutionEvent`.

musketyr:
@sdelamo where is the best place to add such a documentation?

function-aws/src/test/groovy/io/micronaut/function/aws/AfterExecutionEventListener.java
```diff
@@ -0,0 +1,25 @@
+package io.micronaut.function.aws;
+
+import io.micronaut.context.event.ApplicationEventListener;
+import io.micronaut.function.aws.event.AfterExecutionEvent;
+import jakarta.inject.Singleton;
+
+@Singleton
+public class AfterExecutionEventListener implements ApplicationEventListener<AfterExecutionEvent> {
```
sdelamo, line 8: Please, delete it. Each test has its own `static class` annotated with `@Requires(property = 'spec.name'` to avoid bean pollution.

src/main/docs/guide/lambda/requestHandlers.adoc
```diff
@@ -45,3 +45,11 @@ With api:function.aws.MicronautRequestHandler[], it is expected that you supply
 | Raw streams
 | api:function.aws.MicronautRequestStreamHandler[]
 |===
+
+AWS Lambda runtime will reuse the instances of your handler classes but the Java Virtual Machine will be suspended between the invocation. Every call on a single instance of your handler will reuse the same application context. If you need to perform actions before the JVM gets suspended then you can create **synchronous** handler for `AfterExecutionEvent` event which is published right before the execution of the method `handleRequest` in `MicronautRequestHandler` and `MicronautRequestStreamHandler` is finished.
+
+[source,java]
```
sdelamo, line 51: replace with: 

```asciidoc
To subscribe to `AfterExecutionEvent` create a `@Singleton` class which implements `ApplicationEventListener<AfterExecutionEvent>`.  Learn more about https://docs.micronaut.io/latest/guide/#contextEvents[Context Events].
```

sdelamo (changes requested):

src/main/docs/guide/lambda/requestHandlers.adoc
```diff
@@ -45,3 +45,11 @@ With api:function.aws.MicronautRequestHandler[], it is expected that you supply
 | Raw streams
 | api:function.aws.MicronautRequestStreamHandler[]
 |===
+
+AWS Lambda runtime will reuse the instances of your handler classes but the Java Virtual Machine will be suspended between the invocation. Every call on a single instance of your handler will reuse the same application context. If you need to perform actions before the JVM gets suspended then you can create **synchronous** handler for `AfterExecutionEvent` event which is published right before the execution of the method `handleRequest` in `MicronautRequestHandler` and `MicronautRequestStreamHandler` is finished.
```
sdelamo, line 49: Create a new asciidoc file and add it to the `toc.yml`. So that it is easy to link to.

sdelamo (changes requested):

function-aws/src/test/groovy/io/micronaut/function/aws/AfterExecutionEventListener.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-aws/pull/1433/commits/267a5a6c89a08bdaacbe7161f11efd498c3fb3ee

src/main/docs/guide/lambda/requestHandlers.adoc
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-aws/pull/1433/commits/07950ed23101941637007d1158b1dc1abeac96f1

src/main/docs/guide/lambda/requestHandlers.adoc
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-aws/pull/1433/commits/a9f862da9182c451fc7466c4145595e4e8e3689f

sdelamo (approved):

function-aws/src/main/java/io/micronaut/function/aws/MicronautRequestHandler.java
```diff
@@ -143,7 +147,14 @@ public final O handleRequest(I input, Context context) {
         if (!inputType.isInstance(input)) {
             input = convertInput(input);
         }
-        return this.execute(input);
+        try {
+            O output = this.execute(input);
+            resolveAfterExecutionPublisher().publishEvent(AfterExecutionEvent.success(output));
```
graemerocher, line 152: does it make sense to include the `Context` in the event?

musketyr: I'm not sure how much it's worth for the`success` but it could be definitely useful for the`failure`. I will add it.

sdelamo (approved):

PR #1434, opened 2022-08-09
Log context initialization errors

musketyr:
In case of using an external system to monitor errors such as Sentry there is currently no way how to forward context creation errors into the system. 

When there is an issue with the `ApplicationContext` such as missing property then the errors are only logged into CloudWatch which is not sufficient.

Logging the errors in the constructor where they initialization happens is the best way how to keep track of these issues.

### DIFF
--- a/function-aws/src/main/java/io/micronaut/function/aws/MicronautRequestHandler.java
+++ b/function-aws/src/main/java/io/micronaut/function/aws/MicronautRequestHandler.java
@@ -40,11 +40,6 @@ import java.util.Optional;
  */
 public abstract class MicronautRequestHandler<I, O> extends AbstractFunctionExecutor<I, O, Context> implements RequestHandler<I, O>, MicronautLambdaContext {
 
-    /**
-     * Logger for the application context creation errors.
-     */
-    private static final Logger LOG = LoggerFactory.getLogger(MicronautRequestHandler.class);
-
     public static final String ENV_X_AMZN_TRACE_ID = "_X_AMZN_TRACE_ID";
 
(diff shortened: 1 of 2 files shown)


musketyr:
I'm not sure how to resolve the code violations 🤷

graemerocher:
Take a look at the check style report which will provide some hints

musketyr:
I've seen it but it looks to me to report violations which have been there before the PR

musketyr:
@graemerocher fixed

function-aws/src/main/java/io/micronaut/function/aws/MicronautRequestHandler.java
```diff
@@ -93,8 +100,13 @@
      * Lambda deployment.
      */
     public MicronautRequestHandler() {
-        buildApplicationContext(null);
-        injectIntoApplicationContext();
+        try {
+            buildApplicationContext(null);
+            injectIntoApplicationContext();
+        } catch (Exception e) {
+            LOG.error("Exception initializing handler", e);
```
graemerocher, line 107: Perhaps include the message like: 

```java
 LOG.error("Exception initializing handler: " + e.getMessage(), e);
```

graemerocher (changes requested):

function-aws/src/main/java/io/micronaut/function/aws/MicronautRequestHandler.java
musketyr: done

graemerocher:
Thanks for the contribution!

PR #831, opened 2020-12-19
#790 Initial support for aws.*.endpoint-override.

sskrla:

Issue #790, opened 2020-12-08
Support AWS endpoint overrides via configuration values


sskrla:
It is occasionally useful to override the endpoint of AWS clients, especially for testing and Fedramp environments.

Specifically, tests similar to the following would benefit from an endpoint configuration value
```
@MicronautTest
@Testcontainers(disabledWithoutDocker = true)
@TestInstance(PER_CLASS)
internal class S3StorageTest : TestPropertyProvider {
    @Inject
    lateinit var storage: S3Storage

    @Container
    val localstack = LocalStackContainer(LocalStackImage)
        .withServices(S3)
        .also { it.start() }

    @Test fun `stores file`(@TempDir scratch: File) {
        val testFile = File(scratch, "${UUID.randomUUID()}.avro")
        testFile.writeText(randomAscii(1024))
        runBlocking {
            storage.store("test.avro", testFile)
        }
    }

    override fun getProperties() = mapOf(
        "aws.s3.endpoint" to localstack.getEndpointOverride(S3).toString(),
    )
}
```

driverpt:
Any ETA on this one ?

alvarosanchez:
This was done for S3 in #1420

scprek:
Potential fix in https://github.com/micronaut-projects/micronaut-aws/pull/1456 waiting review. @sskrla

sskrla:
> Potential fix in #1456 waiting review. @sskrla

Looks great, and like you out quite a lot of effort into it, thanks for that!

scprek:
@alvarosanchez @sskrla can close this issue?

PR #1435, opened 2022-08-11
Fix endpoint override for S3Async Client

StephanePaulus:
For the S3Client the override property worked, but not for the S3AsyncClient. This PR should fix that.

Now you will be able to override the S3 endpoint with S3
````
aws:
  s3:
    endpoint-override: http://localhost:4566
````
This can also be linked to https://github.com/micronaut-projects/micronaut-aws/issues/790 and https://github.com/micronaut-projects/micronaut-aws/pull/831


StephanePaulus:
Will make a new PR with correct commit author

PR #1445, opened 2022-09-01
build: use a base module to reduce build duplication

sdelamo:


wetted (approved):

wetted:
> I think we can also do this for the test projects. See my revised PR for MQTT [micronaut-projects/micronaut-mqtt#198](https://github.com/micronaut-projects/micronaut-mqtt/pull/198), if you think that's OK.

I did what I suggested, for your review @sdelamo :)

sdelamo:
> I think we can also do this for the test projects. See my revised PR for MQTT [micronaut-projects/micronaut-mqtt#198](https://github.com/micronaut-projects/micronaut-mqtt/pull/198), if you think that's OK.

Yes, I think that is a good idea.

PR #1454, opened 2022-09-06
Switch to using catalog discovery client

timyates:


aws-distributed-configuration/build.gradle.kts
```diff
@@ -4,6 +4,6 @@ plugins {
 
 dependencies {
     api(project(":aws-common"))
-    api("io.micronaut.discovery:micronaut-discovery-client")
+    api(mn.micronaut.discovery)
```
sdelamo, line 7: I don't understand. 

should not be `artifactId` replacing `-` with `.` and preffixing with `mn.` ?

sdelamo: @melix @timyates  ?

sdelamo (approved):

aws-distributed-configuration/build.gradle.kts
melix: It only depends on how it was written in the catalog in the beginning

timyates: And it's been "discovery" for over a year

https://github.com/micronaut-projects/micronaut-core/blob/3.6.x/gradle/libs.versions.toml#L282

PR #1456, opened 2022-09-07
Support global AWS endpoint overrides via configuration values

scprek:
Resurrecting https://github.com/micronaut-projects/micronaut-aws/pull/831. Not sure if there's an issue doing this at in the parent or client builder like in the S3 ticket https://github.com/micronaut-projects/micronaut-aws/pull/1420

fixes: https://github.com/micronaut-projects/micronaut-aws/issues/790


aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/service/ClientConfigurationProperties.java
```diff
@@ -0,0 +1,29 @@
+package io.micronaut.aws.sdk.v2.service;
+
+import io.micronaut.aws.AWSConfiguration;
+import io.micronaut.core.annotation.Nullable;
+
+import java.net.URI;
+
+public abstract class ClientConfigurationProperties extends AWSConfiguration {
```
scprek, line 8: Moved comments from S3ConfigurationProperties here

alvarosanchez (approved):

alvarosanchez (changes requested):
@scprek thank you very much to move this forward!

Please ensure that `./gradlew check` passes locally. Currently the build failed because there are some new files that need the license header. You can apply it automatically running `./gradlew spotlessApply`.

scprek:
@alvarosanchez I was trying to update the adocs too and somehow use aws.clients.*.endpoint-override. I did this by adding the  @EachProperty annotation. However this then caused the client specific configs to be aws.clients.*.s3.endpoint-override
vs staying aws.clients.s3.endpoint-override

scprek:
Screen Shot 2022-09-08 at 7 31 59 AM
Screen Shot 2022-09-08 at 7 32 08 AM

scprek:
@alvarosanchez Everything I tried wrt .adoc auto documentation with configuration properties didn't work or would require lots of changes. So I manually added which is unfortunate

https://github.com/scprek/micronaut-aws/blob/eade027df144890056960a6fb51c45c08104fe62/src/main/docs/guide/sdkv2/serviceClients.adoc

scprek:
Screen Shot 2022-09-12 at 9 59 08 AM

scprek:
Screen Shot 2022-09-12 at 10 00 45 AM

scprek:
I also noticed that there's a config file for secrets manager in a different directory... so now we have configurations in different places https://github.com/micronaut-projects/micronaut-aws/blob/master/aws-secretsmanager/src/main/java/io/micronaut/aws/secretsmanager/SecretsManagerConfigurationProperties.java

scprek:
I seem to have failed `aws-sdk-v2:japiCmp` and not sure how to complete this?
Screen Shot 2022-09-12 at 3 58 41 PM

scprek:
@n0tl3ss @melix @alvarosanchez I made changes to use `@EachProperty` for common configurations

```java
@EachProperty("services")
public class AWSServiceConfiguration extends AWSConfiguration {
```

and one for service specific configs to extend
```java
@ConfigurationProperties(AWSServiceConfigurationProperties.SERVICE_PREFIX)
public class AWSServiceConfigurationProperties extends AWSConfiguration {
```

and would love some feedback. There are currently various spots in the docs for different AWS services(secrets manager, cloudwatchlogging, which I moved?) and them also being nested under sdk2?

Screen Shot 2022-09-13 at 8 56 57 PM
Screen Shot 2022-09-13 at 8 56 52 PM

aws-sdk-v2/src/test/java/io/micronaut/aws/sdk/v2/RekognitionConfigurationProperties.java
```diff
@@ -0,0 +1,24 @@
...
+
+import io.micronaut.context.annotation.ConfigurationBuilder;
+import io.micronaut.context.annotation.ConfigurationProperties;
+import software.amazon.awssdk.services.rekognition.RekognitionClient;
+import software.amazon.awssdk.services.rekognition.RekognitionClientBuilder;
+import software.amazon.awssdk.services.s3.S3Configuration;
+
+//tag::class[]
+@ConfigurationProperties(RekognitionClient.SERVICE_NAME)
+public class RekognitionConfigurationProperties  {
```
scprek, line 12: Most ConfigurationProperties classes I made are empty since the `@EachProperty` change, I did include what was done in the S3ConfigurationProperties. Not sure if every single service needs one of these would have Builders

aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/service/cloudwatchlogs/CloudWatchLogsConfigurationProperties.java
```diff
@@ -0,0 +1,31 @@
...
+import software.amazon.awssdk.services.cloudwatchlogs.CloudWatchLogsClient;
+
+/**
+ * Configuration properties for {@code CloudWatchLogsClient}.
+ *
+ * @author Stephen Cprek
+ * @since 3.9.0
+ */
+@ConfigurationProperties(CloudWatchLogsClient.SERVICE_NAME)
+public class CloudWatchLogsConfigurationProperties extends AWSServiceConfigurationProperties {
```
scprek, line 30: These are currently unused now that I changed over to `@EachProperty`, not sure if they all need builders

scprek:
@alvarosanchez not sure how to address the build issues. I've rebased to master and have run `./gradlew check` along with individual tasks that said they failed locally and everything passes using java 11. Do I have to recompile against java 8 and test locally or is there some trick to run locally?


Sonar failure
```
* What went wrong:
[1462](https://github.com/micronaut-projects/micronaut-aws/actions/runs/3049664265/jobs/4918538490#step:7:1463)
Execution failed for task ':sonarqube'.
[1463](https://github.com/micronaut-projects/micronaut-aws/actions/runs/3049664265/jobs/4918538490#step:7:1464)
> Project not found. Please check the 'sonar.projectKey' and 'sonar.organization' properties, the 'SONAR_TOKEN' environment variable, or contact the project administrator
```

melix:
AFAIK Sonar won't run on community PRs.

alvarosanchez:
> AFAIK Sonar won't run on community PRs.

That's correct.

However, @scprek there are still other failures:

```
FAILURE: Executed 120 tests in 1m 33s (102 failed)
> Task :aws-sdk-v2:test FAILED
```

Note that I still have not reviewed in detail this PR.

scprek:
Understood but I'm unable to recreate those errors locally. I was asking if there is some micronaut build trick for testing against Java 8. As that's what's failing right?

alvarosanchez:
Simply use a JDK 8 locally

scprek:
> Simply use a JDK 8 locally

🤣 

Thanks, I wasn't sure if there was some target flag that needed to change somewhere.

scprek:
Fixed the java 8 issue https://bugs.openjdk.org/browse/JDK-8141508

scprek:
@alvarosanchez I updated to since 3.10. I'm not sure if having the since is required/expected or not.

scprek:
@alvarosanchez is there anyone else I should ask to review?

alvarosanchez:
No need to. Please be patient, it's on my list, and I will get to this soon.

scprek:
Just checking in @alvarosanchez

alvarosanchez:
@scprek I appreciate your patience with this, the notifications about this got buried within a bunch of others. I'm reviewing this now

aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/service/AwsClientFactory.java
```diff
@@ -60,10 +68,13 @@ protected AwsClientFactory(AwsCredentialsProviderChain credentialsProvider, AwsR
      * @return The sync builder
      */
     public SB syncBuilder(SdkHttpClient httpClient) {
-        return createSyncBuilder()
+        SB sb = createSyncBuilder()
                 .httpClient(httpClient)
                 .region(regionProvider.getRegion())
                 .credentialsProvider(credentialsProvider);
+        // Do not replace with method reference - https://bugs.openjdk.org/browse/JDK-8141508
```
alvarosanchez, line 75:
```suggestion
        // Do not replace with method reference - https://bugs.openjdk.org/browse/JDK-8141508
        //noinspection Convert2MethodRef        
```

So that, at least IDEA, stops highlighting the line.

Please add it to the other places too

aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/service/cloudwatchlogs/CloudWatchLogsConfigurationProperties.java
alvarosanchez: Yes, no need to have them as far as I can tell. Please remove.

aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/service/AwsClientFactory.java
```diff
@@ -39,16 +42,21 @@
 
     protected final AwsCredentialsProviderChain credentialsProvider;
     protected final AwsRegionProviderChain regionProvider;
+    @Nullable
+    protected final AWSServiceConfiguration configuration;
```
alvarosanchez, line 46: If it's nullable then it shouldn't be final no?

aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/service/cloudwatchlogs/CloudwatchLogsClientFactory.java
```diff
@@ -45,9 +47,11 @@ public class CloudwatchLogsClientFactory extends AwsClientFactory<CloudWatchLogs
...
      * @param credentialsProvider The credentials provider
      * @param regionProvider      The region provider
+     * @param context             The application context
      */
-    protected CloudwatchLogsClientFactory(AwsCredentialsProviderChain credentialsProvider, AwsRegionProviderChain regionProvider) {
-        super(credentialsProvider, regionProvider);
+    protected CloudwatchLogsClientFactory(AwsCredentialsProviderChain credentialsProvider, AwsRegionProviderChain regionProvider, ApplicationContext context) {
+        super(credentialsProvider, regionProvider, context.findBean(
+            AWSServiceConfiguration.class, Qualifiers.byName(CloudWatchLogsClient.SERVICE_NAME)).orElse(null));
     }
```
alvarosanchez, line 55: I believe you can receive directly the `@Named(CloudWatchLogsClient.SERVICE_NAME) AWSServiceConfiguration` instead of the application context

aws-sdk-v2/src/test/groovy/io/micronaut/aws/sdk/v2/service/CloudWatchLogsClientSpec.groovy
```diff
@@ -1,24 +1,40 @@
...
     void "it can configure an async client"() {
         when:
         CloudWatchLogsAsyncClient client = applicationContext.getBean(CloudWatchLogsAsyncClient)
 
         then:
-        client.serviceName() == CloudWatchLogsClient.SERVICE_NAME
+        client.serviceName() == serviceName()
+    }
+
+    void "it can override the endpoint"() {
```
alvarosanchez, line 31: Perhaps you could move this test to `ServiceClientSpec` and add abstract methods for the subclasses to provide the actual `SdkClient` client and asyncClient instances.

aws-sdk-v2/src/test/java/io/micronaut/aws/sdk/v2/RekognitionConfigurationProperties.java
alvarosanchez: So if they leverage underlying SDK builders, then I think it is justified to have them. Otherwise it's pointless to keep them, IMO

aws-secretsmanager/src/test/groovy/io/micronaut/aws/secretsmanager/SecretsManagerConfigurationSpec.groovy
```diff
@@ -5,7 +5,7 @@ class SecretsManagerConfigurationSpec  extends ApplicationContextSpecification {
     @Override
     Map<String, Object> getConfiguration() {
         super.configuration + [
-                'aws.secretsmanager.enabled': false
+                'aws.services.secretsmanager.enabled': false
```
alvarosanchez, line 8: This is a breaking change. I understand the reason why you changed it (agree that consistency is good), but it being a breaking change would require a new major release of micronaut-aws.

I suggest keeping them as they are for now.

alvarosanchez (changes requested):

alvarosanchez (changes requested):

sdelamo:
@scprek thanks for the PR and the patience. I am going to fork your PR and try to implement this in a non breaking change.

aws-sdk-v2/src/test/java/io/micronaut/aws/sdk/v2/RekognitionConfigurationProperties.java
scprek: Sounds good that’s what I was thinking too

scprek:
Thanks so much @alvarosanchez ans @sdelamo . With respect to reverting the services config breaking change, does that mean the docs will need to be updated to have endpoint override mentioned only where it applies vs what I did to make it global?

aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/service/AwsClientFactory.java
alvarosanchez: Resolving as I see that `regionProvider` is in the same situation

aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/service/s3/S3ConfigurationProperties.java
```diff
@@ -50,8 +50,10 @@ public Builder getBuilder() {
     /**
      * @return The endpoint with which the AWS SDK should communicate
      * @since 3.6.2
+     * Use configuration {@literal aws.services.s3.endpoint-override} instead.
```
alvarosanchez, line 53:
```suggestion
     * @deprecated Use configuration {@literal aws.services.s3.endpoint-override} instead.
```

aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/service/s3/S3ConfigurationProperties.java
```diff
@@ -60,7 +62,9 @@ public URI getEndpointOverride() {
      * Provide a URI to override the endpoint with which the AWS SDK should communicate. Optional. Defaults to `null`.
      * @param endpointOverride The endpoint with which the AWS SDK should communicate
      * @since 3.6.2
+     * Use configuration {@literal aws.services.s3.endpoint-override} instead.
```
alvarosanchez, line 65:
```suggestion
     * @deprecated Use configuration {@literal aws.services.s3.endpoint-override} instead.
```

alvarosanchez (changes requested):
Minor issues in Javadoc, but other than that, good to go!

scprek:
I'll generate the adocs and make sure they look ok

scprek:
@sdelamo @alvarosanchez are we ok with the docs just saying this now? Vs before https://github.com/micronaut-projects/micronaut-aws/pull/1456/files#issuecomment-124613464... the breaking change

Do we now have both `aws.s3.endpoint-override` and `aws.services.s3.endpoint-override` for example? The docs are confusing now.

Screen Shot 2022-12-15 at 11 03 34 AM

alvarosanchez:
Can we not get rid of the `services` prefix altogether?

scprek:
> Can we not get rid of the `services` prefix altogether?

Yeah So I added the s3 prefix so I could have the adocs auto generate it for each service. But since we dropped that, wondering if we should revert back to old paths and just have manual docs.

sdelamo:
> Do we now have both `aws.s3.endpoint-override` and `aws.services.s3.endpoint-override` for example? The docs are can be confusing now

Unfortunately, we have to keep `aws.s3.endpoint-override` as a valid option. Otherwise, we will break users' applications .

sdelamo:
> > Can we not get rid of the `services` prefix altogether?
> 
> Yeah So I added the service prefix so I could have the adocs auto generate it for each service. But since we dropped that, wondering if we should revert back to old paths and just have manual docs.

I think it is better to use `aws.services`. We have many places where we already have configuration with `aws.`. E.g. `aws.client`.
https://micronaut-projects.github.io/micronaut-aws/latest/guide/configurationreference.html

I think it is safer and easier to reason about if we stick with `@EachProperty("aws.services") instead of `@EachProperty("aws")`. If we used @EachProperty("aws")` It will create a bean `AwsServiceConfiguration` with name qualifier `client` and we don't want that.

scprek:
Yeah I like the `aws.services` for those reasons as well. I went ahead and just added a clarifying note to the s3 docs for now.

Screen Shot 2022-12-15 at 11 26 41 AM

scprek:
@sdelamo @alvarosanchez good to go?

alvarosanchez (approved):

melix:
Is this a breaking change which should go in `4.0.0` instead?

scprek:
> Is this a breaking change which should go in `4.0.0` instead?

I believe @sdelamo did work to make it non-breaking and moved that to this PR correct? https://github.com/micronaut-projects/micronaut-aws/pull/1531

Currently for s3 we support both `aws.s3.endpoint-override` and `aws.services.*.endpoint-override` and I think the constructors are backwards compatible.

melix:
Ok great!

PR #1478, opened 2022-10-07
feat: add user agent

sdelamo:


melix (approved):
Instead of a properties file, you may want to generate a class, similar to what is done in Micronaut Test Resources: https://github.com/micronaut-projects/micronaut-test-resources/blob/aca8e883c081514d16a5fdf543165d4d2087c2c9/buildSrc/src/main/groovy/io/micronaut/internal/WriteVersion.java#L32

It is also possible for us to move that feature to Micronaut Build, so that other projects can do that too.

sdelamo:
> It is also possible for us to move that feature to Micronaut Build, so that other projects can do that too.

+ 1 for having this on Micronaut Build. I can see we may need this functionality in other modules (Micronaut GCP, Micronaut Azure). Sending a User agent can help cloud vendors gather information about Micronaut Framework usage in their platforms. 

Can we move ahead with this PR as it currently is? Once Micronaut Build is ready, I will remove the custom build task.

melix:
To be clear, when I talk about moving to Micronaut Build, I'm talking about generating a class which can be used to get the version.

> Can we move ahead with this PR as it currently is? Once Micronaut Build is ready, I will remove the custom build task.

Yes.

sdelamo:
> To be clear, when I talk about moving to Micronaut Build, I'm talking about generating a class which can be used to get the version.

Ok.

aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/SdkClientBuilderListener.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * {@link SdkClientBuilder} listener which configures the User-Agent.
+ *
+ * @author Sergio del Amo
+ * @since 3.10.0
+ */
+@Requires(classes = SdkClientBuilder.class)
+@Singleton
+public class SdkClientBuilderListener implements BeanCreatedEventListener<SdkClientBuilder<?, ?>> {
```
alvarosanchez, line 36: I don't think we should be using a listener for this when it's us creating the bean in the first place. This should be done in the factories directly.

Also, I vaguely remember an issue related with having multiple listeners where one listener would hide the others. @jameskleeh added a warning message, but can't remember what was it.

aws-ua/src/main/java/io/micronaut/aws/ua/UserAgentUtils.java
```diff
@@ -0,0 +1,44 @@
...
+
+    private UserAgentUtils() {
+    }
+
+    /**
+     *
+     * @return User agent value. For example micronaut/3.7.1
+     */
+    @NonNull
+    public static String userAgent() {
```
alvarosanchez, line 39: Would it be useful for users to customise this value? If so, how can they do it? Can they opt out of this?

aws-ua/src/main/java/io/micronaut/aws/ua/VersionInfo.java
```diff
@@ -0,0 +1,52 @@
...
+@Internal
+public class VersionInfo {
+
+    private static final Properties VERSIONS = new Properties();
+
+    static {
+        URL resource = VersionInfo.class.getResource("/micronaut-versions.properties");
+        if (resource != null) {
+            try (Reader reader = new InputStreamReader(resource.openStream(), StandardCharsets.UTF_8)) {
+                VERSIONS.load(reader);
```
alvarosanchez, line 41: This would be adding some overhead to the startup time. Is there an alternative? For example, writing an annotation processor, or using JavaPoet to generate a source file.

alvarosanchez (changes requested):

aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/SdkClientBuilderListener.java
sdelamo: I have changed to set it directly in the parent Factory.

aws-ua/src/main/java/io/micronaut/aws/ua/VersionInfo.java
sdelamo: I have changed to use Javapoet.

sdelamo:
@melix and @alvarosanchez can you check this PR again?. I have changed it to use JavaPoet and created an API `UserAgentProvider`

alvarosanchez (approved):

alvarosanchez (changes requested):
Sorry, one more thing: could you please add a sentence somewhere in the docs mentioning this exists?

sdelamo:
> Sorry, one more thing: could you please add a sentence somewhere in the docs mentioning this exists?

This is not an api the users will use in their applications. What do we want to tell them?

alvarosanchez (approved):
I created #1484 with what I meant

alvarosanchez (approved):

PR #1510, opened 2022-11-17
build: removes dependency to Micronaut HTTP Netty

sdelamo:
It forks cookie utility classes to remove dependency on micronaut-http-netty.


function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/cookie/ObjectUtil.java
```diff
@@ -0,0 +1,44 @@
...
+
+import io.micronaut.core.annotation.Internal;
+
+/**
+ * A grab-bag of useful utility methods.
+ *
+ * Forked from Netty.
+ */
+@Internal
+public final class ObjectUtil {
```
graemerocher, line 25: remove this class it is not necessary we already have methods that do this or can be replaced by `Objects.requireNonNull`

graemerocher (changes requested):
Move all the classes into `io.micronaut.function.aws.proxy` package and make all classes package private instead of public

Also need to cleanup the thread local in the DateFormatter on ctx shutdown

function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/cookie/ObjectUtil.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-aws/pull/1510/commits/95806e27850a7e21ec52ad2db99aa1221fb5a839

graemerocher (approved):

PR #1499, opened 2022-11-01
build: use Micronaut Framework 4.0.0-SNAPSHOT

sdelamo:

PR #1527, opened 2022-12-09
Use new RequestLifecycle api

yawkat:
Untested, can't get the snapshot to run locally. Most of the code that I removed seems to be identical with RouteExecutor (or, now, RequestLifecycle), however I may have missed some differences.

### DIFF
--- a/aws-cloudwatch-logging/build.gradle.kts
+++ b/aws-cloudwatch-logging/build.gradle.kts
@@ -7,6 +7,6 @@ dependencies {
     implementation(libs.logback.json.classic)
     api(libs.awssdk.cloudwatchlogs)
     api(mn.micronaut.runtime)
-    api(mn.micronaut.serde.jackson)
+    api(mnSerde.micronaut.serde.jackson)
     testRuntimeOnly(mn.snakeyaml)
 }
--- a/aws-distributed-configuration/build.gradle.kts
+++ b/aws-distributed-configuration/build.gradle.kts
@@ -4,6 +4,6 @@ plugins {
 
 dependencies {
     api(projects.awsCommon)
-    api(mn.micronaut.discovery)
+    api(mn.micronaut.discovery.core)
     testImplementation(mn.micronaut.http.server.netty)
 }
--- a/aws-distributed-configuration/build.gradle.kts
+++ b/aws-distributed-configuration/build.gradle.kts
@@ -4,6 +4,6 @@ plugins {
 
 dependencies {
     api(projects.awsCommon)
-    api(mn.micronaut.discovery.core)
+    api(libs.micronaut.discovery.client)
     testImplementation(mn.micronaut.http.server.netty)
 }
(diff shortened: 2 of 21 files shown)


graemerocher:
@yawkat maybe speak to @melix as to why you can't get the snapshots working. Would prefer that these were verified as passing before merge

melix:
What do you mean by "can't get the snapshots working locally"? From the failing tests it seems to be because we don't have a version of Micronaut Mongodb compatible with Micronaut 4 yet: https://ge.micronaut.io/s/detg73tpbbbcq/tests/:function-aws:test/io.micronaut.function.aws.SquareHandlerSpec/test%20using%20detached%20mock?focused-exception-line=0-0&top-execution=1

yawkat:
@melix i can't get gradle to use the local artifact even if i use mavenLocal(), and if i use the code from here: https://github.com/micronaut-projects/micronaut-core/wiki/Gradle-Build-Tips#building-a-module-against-a-local-version-of-micronaut-core , it fails in the checkBom task before the tests run

melix:
So you mean using a local version of Micronaut core?

yawkat:
yes. this pr uses new apis

melix:
@yawkat I have updated your PR to use the latest build plugins. By doing this, it will use the snapshot versions of Micronaut from Sonatype snapshot repo.

If I apply this diff:

```diff
diff --git a/buildSrc/src/main/groovy/io.micronaut.build.internal.aws-base.gradle b/buildSrc/src/main/groovy/io.micronaut.build.internal.aws-base.gradle
index d69a20fc8..a5026b45d 100644
--- a/buildSrc/src/main/groovy/io.micronaut.build.internal.aws-base.gradle
+++ b/buildSrc/src/main/groovy/io.micronaut.build.internal.aws-base.gradle
@@ -1,4 +1,5 @@
 repositories {
+    mavenLocal()
     mavenCentral()
     maven { url "https://s01.oss.sonatype.org/content/repositories/snapshots/" }
 }
```

Then [it uses my locally built version](https://ge.micronaut.io/s/sch3gj66zcpjs/dependencies?focusedDependency=WzMsOCw1MzUsWzMsOCxbNTM1XV1d&toggled=W1szXSxbMyw4XV0).

yawkat:
thanks. i had also put the mavenLocal in a different place, so that might have been a problem :)

yawkat:
yea there are still test failures. i will look on monday

yawkat:
There was one bug I fixed here: b71ecdb8d3d07121e0084fdfe44f7109e485a20b

And one I fixed in the main route executor PR: https://github.com/micronaut-projects/micronaut-core/pull/8463#issuecomment-1345275784

I've adjusted one failing test in 29ea99d33a1c02fb27444c8bfd22a25e5105aadb because the behavior is intended.

There are still test failures outside the function-aws-api-proxy module, but they seem to be general micronaut core 4.0 compatibility issues, I don't want to fix them all in this PR.

yawkat:
im going to merge this into 4.0.x and continue work in #1499

PR #1531, opened 2022-12-15
BREAKING: remove @Deprecated code

sdelamo:


alvarosanchez:
There is an opportunity window to accommodate the breaking changes of #1456

sdelamo:
> There is an opportunity window to accommodate the breaking changes of #1456

Lets merge this and we can remove the deprecated code which will be generated by #1456

PR #1569, opened 2023-01-09
Fix HTTP 404 when Empty Body on Embedded Jetty

driverpt:


driverpt:
any ETA on this one?

sdelamo:
> The bug is fixed by marking the Jetty HTTP Request as handled
> So if we don't specify a Body, which is typically the case on an HTTP 204 NO CONTENT, Jetty Server will assume that this is an unhandled request and therefor will return HTTP 404
> And according to Jetty's Documentation:
>         // Inform jetty that this request has now been handled
>         baseRequest.setHandled(true);
> We need to inform Jetty that the request has been properly handled
> 
> https://www.eclipse.org/jetty/documentation/jetty-9/index.html#embedding-jetty

PR #1577, opened 2023-01-17
Support multiple prefixes for aws distributed configuration 

marcin-majewski-sonarsource:
I would like to suggest an addition to support multiple prefixes for micronaut AWS distributed configuration. The change is not breaking and the previous remains the same if the prefixes property is not set. 

The need it fulfills is the situation where the configuration has to be read from multiple secrets that do not share a common prefix. 

Let me know what you think.

### DIFF
--- a/aws-distributed-configuration/src/main/java/io/micronaut/aws/distributedconfiguration/AwsDistributedConfiguration.java
+++ b/aws-distributed-configuration/src/main/java/io/micronaut/aws/distributedconfiguration/AwsDistributedConfiguration.java
@@ -17,6 +17,7 @@ package io.micronaut.aws.distributedconfiguration;
 
 import io.micronaut.core.annotation.Experimental;
 import io.micronaut.core.annotation.NonNull;
+import java.util.List;
 
 /**
  * Configuration for Distributed Configuration using AWS services such as AWS Parameter Store or Secrets Manager.
@@ -33,6 +34,13 @@ public interface AwsDistributedConfiguration {
      */
     @NonNull
     String getPrefix();
+
+    /**
+     * @return List of prefixes for AWS Distributed Configuration resources names. If it is non-empty,
+     * {@link AwsDistributedConfiguration#getPrefix()} is not used
+     */
+    @NonNull
+    List<String> getPrefixes();
 
     /**
      * Delimiter after prefix and application name. For /config/application_dev/micronaut.security.oauth2.clients.mycompanyauth.client-secret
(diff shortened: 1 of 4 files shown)


alvarosanchez (approved):

marcin-majewski-sonarsource:
@alvarosanchez do you know by any chance how can I proceed here? I can see that the sonar scan is failing due to the access to the SONAR_TOKEN secrets, I am pretty sure it is due to the fact, that secret are not propagated when running the action in the PR from the forked repository (looks to be a similar problem here: https://community.sonarsource.com/t/sonar-cannot-be-run-on-pr-from-a-fork/69229 and there are some workaround to be incorporated)

alvarosanchez:
@marcin-majewski-sonarsource don't worry, it's a known issue caused by what you linked. It's an optional check, nevertheless, so as long as the other check pass, it's fine.

I will proceed with merging, but there's one more thing I would like you to change.

aws-distributed-configuration/src/main/java/io/micronaut/aws/distributedconfiguration/AwsDistributedConfiguration.java
```diff
@@ -34,6 +35,13 @@ public interface AwsDistributedConfiguration {
     @NonNull
     String getPrefix();
 
+    /**
+     * @return List of prefixes for AWS Distributed Configuration resources names. If it is non-empty,
+     * {@link AwsDistributedConfiguration#getPrefix()} is not used
+     */
+    @NonNull
+    List<String> getPrefixes();
```
alvarosanchez, line 43: Please add `since` tag pointing to `3.12.1`, and make it a `default` method that returns an empty list.

aws-distributed-configuration/src/main/java/io/micronaut/aws/distributedconfiguration/AwsDistributedConfigurationProperties.java
```diff
@@ -51,6 +53,8 @@ public class AwsDistributedConfigurationProperties implements AwsDistributedConf
     @NonNull
     String prefix = DEFAULT_PREFIX;
 
+    private List<String> prefixes = new ArrayList<>();
```
alvarosanchez, line 56: Annotate with `@NonNull`

aws-distributed-configuration/src/main/java/io/micronaut/aws/distributedconfiguration/AwsDistributedConfiguration.java
```diff
@@ -34,6 +35,13 @@ public interface AwsDistributedConfiguration {
     @NonNull
     String getPrefix();
```
alvarosanchez, line 36: Please, also mention in `getPrefix()` Javadoc that it will not be used if prefixes is set.

alvarosanchez (changes requested):

marcin-majewski-sonarsource:
@alvarosanchez my latest commit acknowledges the comments (I left the comments opened and will let you resolve, not sure what the process is here)

PR #1578, opened 2023-01-19
handle SingleValueHeaders values

n0tl3ss:
This PR handles `SingleValueHeaders` from `AwsProxyRequest`. Those values where never handled. It populates instance of `io.micronaut.http.HttpRequest` Headers with values from `SingleValueHeaders`  I added test that covers that scenario.


function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MicronautAwsProxyRequest.java
```diff
@@ -475,12 +506,24 @@ public <T> Optional<T> get(CharSequence name, ArgumentConversionContext<T> conve
      */
     private class AwsHeaders implements HttpHeaders {
 
-        private Headers multiValueHeaders = awsProxyRequest.getMultiValueHeaders();
+        private final Headers multiValueHeaders = awsProxyRequest.getMultiValueHeaders();
```
sdelamo, line 509: I think we could resolve the complexity of the class by resolving the headers in the constructor.

function-aws-api-proxy-test/src/main/java/io/micronaut/function/aws/proxy/test/DefaultServletToAwsProxyResponseAdapter.java
```diff
@@ -42,12 +43,17 @@ public void handle(@NonNull HttpServletRequest request,
                        @NonNull AwsProxyResponse awsProxyResponse,
                        @NonNull HttpServletResponse response) throws IOException {
         Headers responseHeaders = awsProxyResponse.getMultiValueHeaders();
+        Map<String, String> headers = awsProxyResponse.getHeaders();
```
sdelamo, line 46: it would be nice to extract this into a method.

sdelamo (changes requested):
I have created a PR with suggestions  https://github.com/micronaut-projects/micronaut-aws/pull/1583

sdelamo (approved):

PR #1582, opened 2023-01-27
build: projectAccessors and use version catalogue

sdelamo:


gradle/libs.versions.toml
```diff
@@ -10,11 +16,12 @@ managed-aws-lambda = '1.2.2'
 managed-aws-lambda-events = '3.11.0'
 managed-aws-serverless-core = '1.9.1'
 managed-aws-cdk-lib='2.58.1'
+managed-micronaut-starter = "3.8.0"
```
melix, line 19: issue: potential conflict

I am wondering about this dependency. If feels wrong that a Micronaut project "manages" another one. If multiple projects needs micronaut starter, then it would potentially mean a conflict when building the platform BOM, since we could have different managed versions for the same project.

This should probably be replaced with importing the Micronaut Starter catalog in settings....... once it exists! In the meantime, I think we should probably remove the managed flag for this one.

gradle/libs.versions.toml
```diff
@@ -50,17 +59,24 @@ jackson-afterburner = { module = 'com.fasterxml.jackson.module:jackson-module-af
 jetty-server = { module = 'org.eclipse.jetty:jetty-server', version.ref = 'jetty' }
 jcl-over-slf4j = { module = 'org.slf4j:jcl-over-slf4j', version.ref = 'slf4j' }
 junit-jupiter-engine = { module = 'org.junit.jupiter:junit-jupiter-engine' }
+junit-jupiter-api = { module = 'org.junit.jupiter:junit-jupiter-api' }
 junit-platform-engine = { module = "org.junit.platform:junit-platform-suite-engine" }
+aws-java-sdk-lambda = { module = 'com.amazonaws:aws-java-sdk-lambda' }
```
melix, line 64: thought: I wonder if we should make this one a managed dependency too, even if it's an `implementation` dependency...

melix (changes requested):

gradle/libs.versions.toml
sdelamo: I don't think users add them to their application but I am ok making it a managed dependency.

gradle/libs.versions.toml
sdelamo: I removed the `managed` https://github.com/micronaut-projects/micronaut-aws/pull/1582/commits/873a74b7901a0b5109fc1e1fee63429b09b821ae 

I think the long term solution is that we should treat starter core as a new module.

gradle/libs.versions.toml
melix: Right, let's not make it managed.

melix (approved):

PR #1591, opened 2023-01-31
Fix for drive-by localhost cors deny TCK test

timyates:
When the filter blocks access to localhost from a non-local origin, there is no ROUTE_INFO set on the response object.

Previously, we just called get() on the Optional.  This fails.

I believe the fix is to only continue processing the response if the ROUTE_INFO has been applied to the response.


timyates:
Unsure why we're not running 3.13 against java 8 and 11?

timyates:
Seems to have been done here on the 15th Dec https://github.com/micronaut-projects/micronaut-aws/pull/1530

PR #1595, opened 2023-01-31
[RFC] WIP - api-proxy improvements

driverpt:
- Deprecate AWS Serverless Container
- Add Handlers for multiple payload types
- Bridge everything to Micronaut Http directly
- Add more extension points to Runtime


driverpt:
FYI, the code is still duplicated because it's a Work In Progress

sdelamo:
Is the goal of this PR to support Payload 2 ?

driverpt:
Yes. Payload V2 and ALB Request

sdelamo:
@driverpt Please note that we are refactoring this logic in MN 4. I am not sure we will implement the changes suggested in this PR https://github.com/micronaut-projects/micronaut-aws/pull/1499

sdelamo:
> Yes. Payload V2 and ALB Request.


@driverpt the ideal scenario would be to create separate issues for each which clear steps to reproduce the issues and a sample project.

driverpt:
Agree. But the current implementation is tightly coupled into AWS Serverless Container, will require some rework. I've just pushed more code with my proposal.

driverpt:
@sdelamo , should I proceed with rebasing/updating this PR? Or this will be rejected?

Thanks

driverpt:
No problem. Really appreciate that you were able to reuse most of the code

Issue #1442, opened 2022-08-24
AWS Secret Manager Integration: Allow configurable arbitrary secret prefixes


ChristianMattarVT:
### Feature description

I've been investigating the Micronaut AWS Secret Manager integration. Initial setup works like a charm.
However, in our setup we have several secrets which are shared in various ways among different applications, e.g.
Secret 1 is used by Application A and Application B
Secret 2 is used by Application B and Application C.

The AWS IAM permissions are set up so that each application only has access to secrets that it sees.
If we make the secrets available using a common prefix (e.g. based on commonApplicationName), then, once Micronaut encounters a secret to which it has no access, it will stop fetching due to the exception.

Some ideas for improvements:
- Allow an arbitrarily configurable list of secrets to scan, instead of just based on applicationName and commonApplicationName
- Recover gracefully when iterating through all the listed secrets, and one is encountered where access is denied. This would allow to scan all secrets that are stored and just load those that are accessible.

Thank you!

breader124:
I managed to reproduce the issue described in the first post and will continue working on it

PR #1610, opened 2023-02-12
GH-1442: Handle SecretsManagerException thrown when fetching secrets

breader124:
This PR aims to resolve GH-1442. I implemented the approach based on gracefully handling and logging all issues that might happen when fetching secrets from AWS Secrets Manager. There are little code changes that IMO do the job. However, the proposed solution based on modifying configuration properties to fetch only specified secrets is still an option. Looking forward for your feedback


breader124:
@sdelamo, could you please give it a look?

sdelamo:
Thanks for your contribution.

PR #1619, opened 2023-02-25
build: project Accessors and useStandardizedProjectNames

micronaut-build:
Update common files


.gitignore
```diff
@@ -30,5 +30,3 @@ src/main/docs/resources/css/*.css
 src/main/docs/resources/js/*.js
 src/main/docs/resources/style/*.html
 src/main/docs/resources/img/micronaut-logo-white.svg
-*Generated*.java
-/aws-ua/src/main/java/io/micronaut/aws/ua/GeneratedUserAgentProvider.java
```
timyates, line 34: @sdelamo This file is still generated by the build, and ends up being an untracked file?

aws-ua/build.gradle
```diff
@@ -15,6 +15,9 @@ def deleteUserAgentProviderGenerator = tasks.register("deleteUserAgentProviderGe
 }
 clean.dependsOn(deleteUserAgentProviderGenerator)
 
+tasks.spotlessJava.mustRunAfter(userAgentProviderGenerator)
+tasks.spotlessJavaMisc.mustRunAfter(userAgentProviderGenerator)
```
timyates, line 19: @melix this is required as otherwise [gradle fails to configure the task](https://ge.micronaut.io/s/fvn527exhmubw/failure#1) I assume this is an improvement in Gradle 8?

melix: I guess so, they are always adding new validations. That said, [the fix is wrong](https://melix.github.io/blog/2021/10/gradle-quickie-dependson.html). Instead of the `userAgentProviderGenerator` writing its output to the main sources (uh!!!!), it should use its own directory and declare it as a source directory, so that plugins which perform validation like spotless can check it without explicit dependencies:

```gradle
def userAgentProviderGenerator = tasks.register("userAgentProviderGenerator", io.micronaut.internal.aws.tasks.UserAgentProviderGeneratorTask) {
    version.set(micronautVersion)
    outputDirectory.set(layout.buildDirectory.dir("generated/user-agent"))
}
sourceSets.main.java.srcDir(userAgentProviderGenerator)
```

timyates: Comme ça? https://github.com/micronaut-projects/micronaut-aws/pull/1619/commits/31c8caaf2f9d36e0ccfd35fc7ad6bcc592d8d980

.gitignore
timyates: Actually, this is fixed by https://github.com/micronaut-projects/micronaut-aws/pull/1619/commits/31c8caaf2f9d36e0ccfd35fc7ad6bcc592d8d980

timyates:
we were too slow... micronaut-build force pushed the broken version again 😿

sdelamo (approved):

PR #1650, opened 2023-03-17
Migrate Lambda to use servlet core

timyates:
- It updates to the latest milestones
- It deletes CDK module (it will be moved to starter). 
- It moves away from [AWS Serverless Java Container](https://github.com/awslabs/aws-serverless-java-container). It use servlet core. 
- It supports both payload versions v1 and v2.
- Handlers use standard [AWS Lambda Java Events](https://github.com/aws/aws-lambda-java-libs/tree/main/aws-lambda-java-events) POJOs.
- TCK passes almost complety. 
- There is some work still to do in the the proxy-test. Several tests don't pass.


timyates:
Flux test currently fails with `Expected received body of '{"scanAvailable":true,"prefetch":-1}' to contain '[{"name":"Joe"},{"name":"Lewis"}]'` so it's serializing the Flux, not the contents of the flux... I've seen this before, I wonder what I'm missing.

This also doesn't currently support binary payloads in the body, we will need to mime encode them

graemerocher:
do you have the `micronaut-reactor` module on the classpath for the tests?

aws-cloudwatch-logging/build.gradle.kts
```diff
@@ -4,9 +4,14 @@ plugins {
...
 dependencies {
     api(projects.micronautAwsSdkV2)
-    implementation(libs.logback.json.classic)
     api(libs.awssdk.cloudwatchlogs)
     api(mn.micronaut.runtime)
     api(mnSerde.micronaut.serde.jackson)
+
+    implementation(mnLogging.logback.classic)
+    implementation(libs.logback.json.classic) {
+        exclude(group = "ch.qos.logback")
```
graemerocher, line 13: I think you will need explicit excludes for each module for this to work for Maven

function-aws-api-proxy-test/src/test/groovy/io/micronaut/function/aws/proxy/test/AwsApiProxyTestServerSpec.groovy
```diff
@@ -31,6 +32,7 @@ class AwsApiProxyTestServerSpec extends Specification {
         result == 'good'
     }
 
+    @PendingFeature
```
sdelamo, line 35: this test should pass

function-aws-api-proxy-test/src/test/groovy/io/micronaut/function/aws/proxy/test/AwsApiProxyTestServerSpec.groovy
```diff
@@ -60,6 +60,7 @@ class AwsApiProxyTestServerSpec extends Specification {
         result == 'get:prebar+postbar'
     }
 
+    @PendingFeature
```
timyates: This passes for me, so I removed the PendingFeature

timyates:
So now form encoded seems to work, but BodyTest is failing on reactive types 🤔

aws-cloudwatch-logging/build.gradle.kts
timyates: Do you have an example? 🤔

timyates:
@graemerocher done in https://github.com/micronaut-projects/micronaut-aws/pull/1650/commits/64cbf99c29c287ebb4eef13a30a36a45422648b0

@sdelamo is this ok?  I'm not 100% when those classes are used, or what the expectation with exception handling is...

function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/payload1/ApiGatewayProxyRequestEventFunction.java
```diff
@@ -0,0 +1,58 @@
...
+
+    public ApiGatewayProxyRequestEventFunction(ApplicationContext ctx) {
+        super(ctx);
+        startThis(applicationContext);
+        httpHandler = initializeHandler();
+    }
+
+    private ServletHttpHandler<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent> initializeHandler() {
+        ApiGatewayProxyEventHandler apiGatewayProxyEventHandler = new ApiGatewayProxyEventHandler(applicationContext);
+        Runtime.getRuntime().addShutdownHook(
```
driverpt, line 47: Suggestion: Add Base class `AWSLambdaServletHandler` w/ method `onShutdown()`

aws-cloudwatch-logging/build.gradle.kts
graemerocher: not sure what you mean, you need to specify both the group and the module id

function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MultiValue.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+/**
+ *
+ */
+public class MultiValue implements ConvertibleMultiValues<String> {
```
graemerocher, line 36: add `@internal`, give this type a better name, if it can be package private better

graemerocher (changes requested):
looks good, just go through the code and annotate more types and packages with `@Internal` since I think we should keep the public API surface area here small

Issue #1687, opened 2023-04-18
Cloudwatch logger appender doesn't work with native image


andriy-dmytruk:
## Reproduce

* Create a micronaut application ([micronaut launch link](https://micronaut.io/launch?type=DEFAULT&name=demo&package=com.example&javaVersion=JDK_17&lang=JAVA&build=GRADLE&test=JUNIT&version=3.9.0)).
* Add the following dependency:
  ```
  implementation("io.micronaut.aws:micronaut-aws-cloudwatch-logging")
  ```
* Modify the `src/main/resources/logback.xml` to be the following:
  ```xml
  <configuration>
    <appender name='CLOUDWATCH' class='io.micronaut.aws.cloudwatch.logging.CloudWatchLoggingAppender'>
        <groupName>MyLogGroup</groupName>
        <streamName>MyStreamName</streamName>
        <createGroupAndStream>true</createGroupAndStream>
        <encoder class='ch.qos.logback.core.encoder.LayoutWrappingEncoder'>
            <layout class='ch.qos.logback.contrib.json.classic.JsonLayout'>
                <jsonFormatter class='io.micronaut.aws.cloudwatch.logging.CloudWatchJsonFormatter'/>
            </layout>
        </encoder>
    </appender>

    <root level='INFO'>
        <appender-ref ref='CLOUDWATCH'/>
    </root>
  </configuration>
  ```
* Run `./gradlew nativeRun`

## Error

With the configuration in `build.gradle`:
```
graalvmNative {
    binaries.configureEach {
        buildArgs.addAll(
                "--trace-object-instantiation=java.lang.Thread"
        )
    }
}
```
the following exception is thrown (important lines marked):
```
Error: Detected a started Thread in the image heap. Threads running in the image generator are no longer running at image runtime.  Object has been initialized by the io.micronaut.runtime.Micronaut class initializer with a trace: 
 	at java.lang.Thread.<init>(Thread.java:708)
	at java.util.concurrent.Executors$DefaultThreadFactory.newThread(Executors.java:660)
	at ch.qos.logback.core.util.ExecutorServiceUtil$1.newThread(ExecutorServiceUtil.java:42)
	at java.util.concurrent.ThreadPoolExecutor$Worker.<init>(ThreadPoolExecutor.java:630)
	at java.util.concurrent.ThreadPoolExecutor.addWorker(ThreadPoolExecutor.java:920)
[...]

PR #1688, opened 2023-04-18
#1687 Fix native image build with CloudWatchLoggingAppender

andriy-dmytruk:
- Postpone the dispatch thread creation to runtime. Start the thread to send logs to cloudwatch only when `append` method is called, instead of `start`. This way the thread is not started during build time, which is unsupported by native image. (See #1687 for details).
- Create `dispatchOnStart` property to return to previous behavior if needed.


graemerocher (approved):

n0tl3ss (approved):

andriy-dmytruk:
@graemerocher Can we merge it?

PR #1708, opened 2023-05-04
Graal

sdelamo:


function-aws-api-proxy/src/main/resources/META-INF/native-image/io.micronaut.aws/micronaut-function-aws-api-proxy/native-image.properties
```diff
@@ -1 +0,0 @@
-Args = --initialize-at-run-time=io.micronaut.function.aws.proxy --initialize-at-build-time=io.micronaut.function.aws.proxy.model.factory.MicronautAwsProxyResponseFactory
```
sdelamo, line 1: If I don't remove this, just by adding the dependency `micronaut-function-aws-api-proxy`, I was getting: 

```
Error: Classes that should be initialized at run time got initialized during image building:
 org.slf4j.LoggerFactory was unintentionally initialized at build time. To see why org.slf4j.LoggerFactory got initialized use --trace-class-initialization=org.slf4j.LoggerFactory
To see how the classes got initialized, use --trace-class-initialization=org.slf4j.LoggerFactory
Error: Use -H:+ReportExceptionStackTraces to print stacktrace of underlying exception
```

graemerocher (approved):

sdelamo:
Thanks for the help @graemerocher

PR #1757, opened 2023-06-07
graal: more TypeHint.AccessTypes for Lambda Events

sdelamo:
To solve errors when deploying a GraalVM Native Image with jackson databind to a Lambda custom runtime. 

```
io.micronaut.http.client.exceptions.HttpClientResponseException: 
Error decoding HTTP response body: 
    Error decoding JSON stream for type [APIGatewayProxyRequestEvent]: 
Cannot construct instance of `com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent`: 
  cannot deserialize from Object value (no delegate- or property-based Creator): this appears to be a native image, in which case you may need to configure reflection for the class that is to be deserialized

```

### DIFF
--- a/function-aws-custom-runtime/build.gradle.kts
+++ b/function-aws-custom-runtime/build.gradle.kts
@@ -3,6 +3,7 @@ plugins {
 }
 
 dependencies {
+    annotationProcessor(mn.micronaut.graal)
     compileOnly(projects.micronautFunctionAwsApiProxy)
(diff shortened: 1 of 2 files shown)


timyates (approved):
Would be good to have a test-suite-native so this can be checked in this module?

sdelamo:
> Would be good to have a test-suite-native so this can be checked in this module?

Yes, we have to move native test into this module

Issue #1733, opened 2023-04-26
Support for running AWS Lambdas under ALB


johncurrier:
I'm attempting to convert my RESTful AWS Lambda from API Gateway to Elastic Load Balancer (ELB). I got it somewhat working, then realized that the responses were always coming back with a content type of `application/octet-stream`. In trying to figure out why I discovered that the Lambda is returning responses that are transformed to match API Gateway requirements.

There doesn't seem to be a mechanism available to disable that API Gateway-specific transformation of responses. This ticket is to request a Lambda handler that doesn't do that transformation. Ideally it'd support single-value headers.

### Steps to Reproduce

Documented at https://github.com/johncurrier/micronautLambdaUnderAlb.

### Expected Behaviour

A standard RESTful response with the headers set by a controller returned to clients.

### Actual Behaviour

AWS API Gateway-based responses are returned instead. Things like content type come back as their default `application/octet-stream`.

### Environment Information

- **Operating System**: N/A
- **Micronaut Version:** 3.9.0
- **JDK Version:** 11

### Example Application

- https://github.com/johncurrier/micronautLambdaUnderAlb

sdelamo:
@johncurrier we added in Micronaut Framework 4 a handler for ALB. Can you try https://github.com/johncurrier/micronautLambdaUnderAlb/pull/1 and confirm this is solved?

johncurrier:
@sdelamo it all looks promising, but I've given up on Micronaut for the set of projects that I'm working on.

sdelamo:
> @sdelamo it all looks promising, but I've given up on Micronaut for the set of projects that I'm working on.

due to use not supporting alb event?

johncurrier:
@sdelamo it's just too difficult to try to figure out how to work around the various flaws in how it all works. Some of the fundamental pieces seem to be singletons, so having a Lambda that interacts with multiple external services requires way too much knowledge (and hacks) related to the underpinnings. I also ran into period connection failures that all disappeared when I moved to a different container.

I've always loved Micronaut, but it's just too difficult to get anything beyond simple things working.

PR #1765, opened 2023-06-12
Adds Application Load Balancer Handler

sdelamo:
see: https://github.com/micronaut-projects/micronaut-aws/issues/1733


timyates (approved):
Have we tested this to make sure it fixes the issue [they describe in their verification](https://github.com/johncurrier/micronautLambdaUnderAlb#verification)?

sdelamo:
> Have we tested this to make sure it fixes the issue [they describe in their verification](https://github.com/johncurrier/micronautLambdaUnderAlb#verification)?

No, I am going to do that.

PR #1778, opened 2023-06-16
Fix the AWS tck-function-aws-api-proxy-test tests

timyates:
We had two issues here.

1. We needed to switch off the AWS health status (as with the other TCK tests)
    To do this, we use a property in a yaml file
2. When we ran the tests we have 2 application contexts.
    One of them is in the Lambda handler -- this is the one we want
    The other is in the test-server -- this is the one we don't want
    Tests were failing as we were using beans from one in the test, and then beans from the other in the assertion.
    Fixed by returning the handler application context from the test-server


sdelamo (approved):
Great to see the tests green

PR #1791, opened 2023-06-30
Enable binary compatibility check

wetted:


wetted:
@Sergio I re-enabled binary compatibility check and set the baselineVersion to latest milestone. However, the check fails. How do you want to handle this? Defer re-enabling, it, add the changes to a`ccepted-api-changes.json`, or perhaps a new milestone release as the baseline will resolve it?

sdelamo (approved):

PR #1852, opened 2023-08-29
accept api changes for serde import introspections

sdelamo:
I have no idea why we need to do this change. we have not changed this file https://github.com/micronaut-projects/micronaut-aws/commits/master/aws-lambda-events-serde/src/main/java/io/micronaut/aws/lambda/events/serde/CognitoUserPoolMigrateUserEventSerde.java


dstepanov:
I wonder if we should annotate the generated classes as `@Internal` to be skipped by japicmp

sdelamo:
> I wonder if we should annotate the generated classes as `@Internal` to be skipped by japicmp

Maybe.

PR #1886, opened 2023-09-19
Introduce mechanism to avoid loading default config files

guillermocalvo:


sdelamo:
@guillermocalvo check the GCP module. we probably have the same situation.

guillermocalvo:
> check the Azure module. we probably have the same situation.

@sdelamo On it 👍

sdelamo:
@guillermocalvo Sorry but I am going to close this PR. I have been thinking more about this and I think with the already existing options users can reduce the number of searched prefixes. see: https://github.com/micronaut-projects/micronaut-aws/pull/2033

PR #1888, opened 2023-09-20
fix: multi value header handling

### LINKED ISSUE #1861: [Bug] API Gateway Payload V2: Multiple Headers with same name come as one
### Expected Behavior

`HttpRequest<?>` to return proper Headers

### Actual Behaviour

Returns same Header with values separated by Commas

### Steps To Reproduce

```
    @Issue('https://github.com/micronaut-projects/micronaut-aws/issues/1858')
    void 'ensure multiple headers are properly received'() {
        when:
        HttpResponse<Integer> response = client.toBlocking()
                .exchange(HttpRequest.GET('/receive-multiple-headers')
                            .header(HttpHeaders.ETAG, "A")
                            .header(HttpHeaders.ETAG, "B"), Integer.class)

        then:
        response.status == HttpStatus.OK
        response.body.get() == 2
    }
```

```
        @Get(value = '/receive-multiple-headers', processes = MediaType.TEXT_PLAIN)
        Integer receiveMultipleHeaders(HttpRequest<?> request) {
            return request.headers.getAll(HttpHeaders.ETAG).size()
        }
```

### Environment Information

- Operating System: MacOS
- JVM: 17

### Example Application

_No response_

### Version

Latest with AWS 4.0.3

sdelamo:
Close #1861

[Multi-header values should use a comma](https://datatracker.ietf.org/doc/html/rfc2616#section-4.2): 

> Multiple message-header fields with the same field-name MAY be
   present in a message if and only if the entire field-value for that
   header field is defined as a comma-separated list [i.e., #(values)].
   It MUST be possible to combine the multiple header fields into one
   "field-name: field-value" pair, without changing the semantics of the
   message, by appending each subsequent field-value to the first, each
   separated by a comma.

But there are headers which have a , without using it to designate a multi-value. [E.g. Date](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Date).  Thus, a mess 🤷🏼‍♂️

### DIFF
--- a/function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MapCollapseUtils.java
+++ b/function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MapCollapseUtils.java
@@ -16,9 +16,13 @@
 package io.micronaut.function.aws.proxy;
 
 import io.micronaut.core.annotation.Internal;
+import io.micronaut.core.annotation.NonNull;
+import io.micronaut.core.annotation.Nullable;
 import io.micronaut.core.util.CollectionUtils;
+import io.micronaut.http.HttpHeaders;
 import io.micronaut.http.MutableHttpHeaders;
 
+import java.util.Arrays;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -30,6 +34,8 @@ import java.util.Map;
  */
 @Internal
 public final class MapCollapseUtils {
+
+    private static final List<String> HEADERS_ALLOWING_COMMAS = Arrays.asList(HttpHeaders.DATE, HttpHeaders.USER_AGENT);
 
     private MapCollapseUtils() {
     }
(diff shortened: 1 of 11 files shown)


sdelamo:
ping @driverpt

function-aws-api-proxy/src/test/groovy/io/micronaut/function/aws/proxy/MapCollapseUtilsSpec.groovy
```diff
@@ -0,0 +1,77 @@
...
+import io.micronaut.core.convert.ConversionService
+import io.micronaut.http.CaseInsensitiveMutableHttpHeaders
+import io.micronaut.http.HttpHeaders
+import io.micronaut.http.MutableHttpHeaders
+import io.micronaut.test.extensions.spock.annotation.MicronautTest
+import jakarta.inject.Inject
+import spock.lang.Specification
+
+@MicronautTest
+class MapCollapseUtilsSpec extends Specification {
```
sdelamo, line 14: The examples `Foo` and `Key` in this test are taken directly from the AWS Lambda Console. I did request and logged them to see how they are represented in the Lambda Event Objects.

function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MapCollapseUtils.java
```diff
@@ -31,6 +35,8 @@
 @Internal
 public final class MapCollapseUtils {
 
+    private static final List<String> HEADERS_ALLOWING_COMMAS = Arrays.asList(HttpHeaders.DATE, HttpHeaders.USER_AGENT);
```
timyates, line 38: Should this be configurable?

timyates (approved):

function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/MapCollapseUtils.java
sdelamo: I don't see how Netty handles this @yawkat do you know?

sdelamo: > Should this be configurable?

but it should be configurable at not in AWS right?

PR #1916, opened 2023-10-16
fix: CustomPojoSerializer only for Serde

### LINKED ISSUE #1914: MicronautRequestHandler no longer working with SQS and Jackson
### Expected Behavior

After bumping to Micronaut 4 we've experienced some weird stuff regarding serialization and deserialization. For reasons, we cannot use Serde in our project (that has been live in production for more than two years). We want to keep depending on Jackson as before. 

I've created a small POC, see "Example Application" below. The basis of the project were created using Micronaut Launch, see [micronaut-cli.yml](https://github.com/thobiaskarlsson/micronaut-4-jackson-sqs-issue/blob/main/micronaut-cli.yml) for launch details.

### Actual Behaviour

When consuming SQS Events on a Lambda, it fails with a `NullPointerException` due to the incoming `SQSEvent` (or intenral content) being `null`. 

### Steps To Reproduce

1. See https://github.com/thobiaskarlsson/micronaut-4-jackson-sqs-issue
2. Deploy this JAR to AWS
3. Connect the `com.example.SqsHandler` as an event handler to an SQS queue as event source.
4. Fire any SQS message on the queue
5. Lambda function fails, with the following exception:
```
Cannot invoke "java.util.List.iterator()" because the return value of "com.amazonaws.services.lambda.runtime.events.SQSEvent.getRecords()" is null: java.lang.NullPointerException
java.lang.NullPointerException: Cannot invoke "java.util.List.iterator()" because the return value of "com.amazonaws.services.lambda.runtime.events.SQSEvent.getRecords()" is null
	at com.example.SqsHandler.execute(SqsHandler.java:29)
	at com.example.SqsHandler.execute(SqsHandler.java:15)
	at io.micronaut.function.aws.MicronautRequestHandler.handleRequest(MicronautRequestHandler.java:110)
```
Please note that we've managed to reproduce where the `SQSEvent` (main request object) was `null`, but it seems to have switched to the `getRecords()` being `null` now instead. Either way, it's very strange imo. 

### Environment Information

- JDK 17

### Example Application

https://github.com/thobiaskarlsson/micronaut-4-jackson-sqs-issue

### Version

4.1.4

sdelamo:
Close #1914

This PR remove  the `CustomPojoSerializer` when using  `JacksonDatabind`.

Jackson’s  `ObjectMapper ` creation requires customization which I don’t think it make sense to replicate in our  `CustomPojoSerializer`.

https://github.com/aws/aws-lambda-java-libs/blob/main/aws-lambda-java-serialization/src/main/java/com/amazonaws/services/lambda/runtime/serialization/factories/JacksonFactory.java#L57-L93

### DIFF
--- a/src/main/docs/guide/lambda/eventsLambdaSerde.adoc
+++ b/src/main/docs/guide/lambda/eventsLambdaSerde.adoc
@@ -2,4 +2,5 @@ If you want to use https://micronaut-projects.github.io/micronaut-serialization/
 
 dependency:micronaut-aws-lambda-events-serde[groupId="io.micronaut.aws"]
 
-`micronaut-aws-lambda-events-serde` provides an implementation of `com.amazonaws.services.lambda.runtime.CustomPojoSerializer` which is loaded via SPI. This `CustomPojoSerialization` avoids your Micronaut function to pay a double hit on performances when using a serialization library inside the Lambda function.
+`micronaut-aws-lambda-events-serde` provides an implementation of `com.amazonaws.services.lambda.runtime.CustomPojoSerializer` which is loaded via SPI.
+This `CustomPojoSerialization` avoids your Micronaut function paying a double hit on performance when using a serialization library inside the Lambda function.


aws-lambda-events-serde/src/main/resources/META-INF/services/com.amazonaws.services.lambda.runtime.CustomPojoSerializer
timyates, line 1: The class this points to is in a different module, which is not imported by this modules build (afaics)

timyates (approved):
1 question and a suggestion for the docs wording...

I [pushed a commit](https://github.com/micronaut-projects/micronaut-aws/pull/1916/commits/311ba814fba9c8b6085250aa3a11ecadbec1ddef) that cleaned up unused imports, added a private constructor to `CustomPojoSerializerUtils` and used generics in `CustomPojoSerializerUtils.serializeFromJson` -- which I think is correct

The reproducer for this issue passes locally, so I assume this only fails on AWS?  Have you tried the reproducer on AWS with both Serde and Jackson?

sdelamo:
> The reproducer for this issue passes locally, so I assume this only fails on AWS? Have you tried the reproducer on AWS with both Serde and Jackson?

It did not passed locally. It failed. 

I moved `src/main/resources/META-INF/services/com.amazonaws.services.lambda.runtime.CustomPojoSerializer` to 
`aws-lambda-events-serde/src/main/resources/META-INF/services/com.amazonaws.services.lambda.runtime.CustomPojoSerializer` so that for Jackson-databind it is not loaded. 

I instead load in our tests the `ObjectMapper` with `LambdaEventSerializers.serializerFor`which is the way internally AWS does it if you don't use have a `CustomPojoSerializer`.

aws-lambda-events-serde/src/main/resources/META-INF/services/com.amazonaws.services.lambda.runtime.CustomPojoSerializer
sdelamo: I think it is fine. I did not want to move the class to avoid any breaking change and also because the class is Micronaut Serialization independent. I am not sure if in the future we can reconsider and use CustomPojoSerializer for Jackson Databind as well.

sdelamo: The way the mixins and Pojos are rewritten in the AWS Java Events Library requires a custom Jackson Object Mapper which I am not sure we can load easily.

Issue #1920, opened 2023-10-17
[Doc] Distributed configuration with SecretsManager


chjourdain:
### Issue description

Hello,

While using micronaut 3, with a distributed configuration on secretManager I declared the following `bootstrap.yml`

```
micronaut:
  config-client:
    enabled: true
  application:
    name: someApplication
aws:
  secretsmanager:
    secrets:
      - secret-name: rds

```

Moving to micronaut 4, I had to add a 'client' wrapper, or it refused to work

```
micronaut:
  config-client:
    enabled: true
  application:
    name: someApplication
aws:
  client:
    secretsmanager:
      secrets:
        - secret-name: rds

```

The [documentation for secretManager](https://micronaut-projects.github.io/micronaut-aws/latest/guide/#distributedconfigurationsecretsmanager)  does not mention it whereas every other item do [ParameterStore_documentation](https://docs.micronaut.io/4.0.2/guide/index.html#distributedConfigurationAwsParameterStore)

## Version

Latest with AWS 4.0.3

sdelamo:
@chjourdain the configuration seems to still be `aws.secretsmanager.secrets[0].secrets-name` `aws.secretsmanager.secrets[0].prefix`. See: https://github.com/micronaut-projects/micronaut-aws/blob/master/aws-secretsmanager/src/main/java/io/micronaut/aws/secretsmanager/SecretsManagerConfigurationProperties.java

How, can I reproduce the error you are describing.

PR #1921, opened 2023-10-17
fix(doc): update secretManager distributed configuration documentatio…

chjourdain:
fix doc (#1920)


sdelamo:
You would need to sign the CLA. Moreover, I am not sure this is correct. I will comment in the related issue.

Issue #1812, opened 2023-07-18
Regression in Lambda Native performance


sdelamo:
### Micronaut Framework 3

<img width="1650" alt="CleanShot 2023-07-18 at 15 58 29@2x" src="https://github.com/micronaut-projects/micronaut-aws/assets/864788/073024a0-870c-4242-82b0-755d116eb4c2">

### Micronaut Framework 4

<img width="1435" alt="CleanShot 2023-07-18 at 14 53 26@2x" src="https://github.com/micronaut-projects/micronaut-aws/assets/864788/e291b67b-54fd-4e5c-8a45-31dc02f6801d">

see cold startup times in the bottom right corner of the images have significantly increased in MN4.

sdelamo:
Using Micronaut Jackson Databind helps 

<img width="1560" alt="CleanShot 2023-07-18 at 18 31 25@2x" src="https://github.com/micronaut-projects/micronaut-aws/assets/864788/00cf6e00-7e78-4de9-a430-1cda02f7fcf2">

timyates:
@sdelamo did https://github.com/micronaut-projects/micronaut-aws/pull/1935 fix it?

sdelamo:
No. unfortunately the regression is still present. 

I run [AWS Lambda Todo](https://github.com/micronaut-projects/micronaut-lambda-todo) load tests. 

For Micronaut Framework 4 (branch `master`). 

| Runtime                | Max Cold Startup ms | 
|:-----------------------|:--------------------|
| Java runtime           | 6849              |
| Native                 | 1131              |
| Java SnapStart         | 2960              |
| Java SnapStart Priming | 2267              |

For Micronaut Framework 4 (branch `micronaut-3`). 

| Runtime                | Max Cold Startup ms | 
|:-----------------------|:--------------------|
| Java runtime           | 8243              |
| Native                 | 922               |
| Java SnapStart         | 3625              |
| Java SnapStart Priming | 3493              |

sdelamo:
This is now fixed since Micronaut Framework 4.3.0

PR #1935, opened 2023-11-02
eagerly intialize singletons for payload v1, v2 and alb handlers

sdelamo:
See: https://github.com/micronaut-projects/micronaut-aws/issues/1812


function-aws-api-proxy/src/test/groovy/io/micronaut/function/aws/proxy/payload1/Payload1EagerlyInitializeSingletonTest.groovy
```diff
@@ -0,0 +1,21 @@
...
+import spock.lang.Specification
+
+import java.time.LocalDateTime
+
+class Payload1EagerlyInitializeSingletonTest extends Specification {
+
+    void "verify singletons are eagerly intialized for Payload v1 function"() {
+        given:
+        ApiGatewayProxyRequestEventFunction handler = new ApiGatewayProxyRequestEventFunction()
+        sleep(5_000)
```
timyates, line 13: Instead of `Clock` and a 5s sleep in the tests, could we have:

```groovy
@Singleton
class SingletonBean {

    static String PACKAGE = null;

    SingletonBean(ApplicationContext ctx) {
        PACKAGE = ctx.environment.packages.find()
    }
}
```

And then this test becomes:

```
class Payload1EagerlyInitializeSingletonTest extends Specification {

    void "verify singletons are eagerly intialized for Payload v1 function"() {
        given:
        ApiGatewayProxyRequestEventFunction handler = new ApiGatewayProxyRequestEventFunction()

        expect:
        SingletonBean.PACKAGE == ApiGatewayProxyRequestEventFunction.packageName

        cleanup:
        handler.close()
    }
}
```

timyates (approved):

PR #1984, opened 2023-12-12
Failing test for issue 1870

timyates:
When annotated with `@MicronautLambdaTest`, this test fails.

It is because the `EagerSingleton` inside the Hander and the Test are 2 different mocks.

When this test is run with `@MicronautTest` then it passes...

It  must be something to do with Environments, as using

```
@MicronautTest(environments = {Environment.FUNCTION, MicronautLambdaContext.ENVIRONMENT_LAMBDA, Environment.TEST})
```

Also fails with the same issue...

### DIFF
--- a/function-aws-test/build.gradle.kts
+++ b/function-aws-test/build.gradle.kts
@@ -9,6 +9,7 @@ dependencies {
 
     testAnnotationProcessor(mn.micronaut.inject.java)
 
+    testImplementation("org.mockito:mockito-core:5.3.1")
     testRuntimeOnly(libs.junit.jupiter.engine)
     testRuntimeOnly(mn.snakeyaml)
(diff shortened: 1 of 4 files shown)


timyates:
Weirdly the same test in Micronaut-test doesn't fail when I change the environments...

So it must be some interaction happening in this aws module...

Found a workaround which is to annotate the field with `@MockBean` instead of `@Inject` as in this test in micronaut-test

https://github.com/micronaut-projects/micronaut-test/blob/2c6b1d198ab362f081a2874e2bc9e28095f1e9ca/test-junit5/src/test/java/io/micronaut/test/junit5/MathMockFieldCollaboratorTest.java#L22-L21

timyates:
Superseded by https://github.com/micronaut-projects/micronaut-aws/pull/2007

PR #2007, opened 2024-01-10
Add a test to verify MockBean works with MicronautLambdaTest

### LINKED ISSUE #1870: MicronautLambdaTest and MockBean don't work as expected
### Expected Behavior

(Raised it also in mictonaut-core, but thought it should probably be here - https://github.com/micronaut-projects/micronaut-core/issues/9708)

When having a MicronautLambdaTest and a MockBean I expect that the MockBean method will be called once on the https://github.com/beforeeach and that mock will be available in my tests, so I could use Mockito's verify methods.
This is the behaviour when using MicronautTest

### Actual Behaviour

While the actual bean is mocked and the regular bean is not being called during the test, it seems that the been is been mocked multiple times, so I don't have one consistent mock I can do verify methods on

### Steps To Reproduce

Lambda:
```java

package com.mimecast.micronaut.lambda;

import com.amazonaws.services.lambda.runtime.events.ScheduledEvent;
import com.mimecast.micronaut.lambda.service.HelloWorldService;
import io.micronaut.context.ApplicationContext;
import io.micronaut.function.aws.MicronautRequestHandler;
import jakarta.inject.Inject;
import lombok.extern.slf4j.Slf4j;

@Slf4j
public class FunctionRequestHandler extends MicronautRequestHandler<ScheduledEvent, Void> {


    @Inject
    private HelloWorldService helloWorldService;

    //used in AWS
    public FunctionRequestHandler() {
    }

    //used in tests
    public FunctionRequestHandler(ApplicationContext applicationContext) {
        super(applicationContext);
    }

    @Override
    public Void execute(ScheduledEvent input) {
        log.info("input: {}", input);
        helloWorldService.handleInvocation(input.getAccount(), input.getTime());
        return null;
    }
}
``` 
Test Class

```java
package com.mimecast.micronaut.lambda;
import com.amazonaws.services.lambda.runtime.events.ScheduledEvent;
import com.mimecast.micronaut.lambda.service.HelloWorldService;
import com.mimecast.micronaut.lambda.service.HelloWorldServiceImpl;
import io.micronaut.context.ApplicationContext;
import io.micronaut.function.aws.test.annotation.MicronautLambdaTest;
import io.micronaut.test.annotation.MockBean;
import jakarta.inject.Inject;
import org.joda.time.DateTime;
import org.junit.jupiter.api.Test;

import static org.mockito.Mockito.mock;
import static org.mockito.Mockito.verify;


@MicronautLambdaTest
class FunctionRequestHandlerTest {
    @Inject
    private ApplicationContext context;

    @Inject
    private HelloWorldService helloWorldService;

    @Test
    void testHandler(){
        FunctionRequestHandler functionRequestHandler = new FunctionRequestHandler(context);
        ScheduledEvent scheduledEvent = new ScheduledEvent();
        String account = "AccountA";
        DateTime dateTime = DateTime.now();
        scheduledEvent.setAccount(account);
        scheduledEvent.setTime(dateTime);
        functionRequestHandler.execute(scheduledEvent);
        verify(helloWorldService).handleInvocation(account, dateTime);
    }

    @MockBean(HelloWorldServiceImpl.class)
    HelloWorldService helloWorldService() {
        return mock(HelloWorldService.class);
    }
}

``` 

This fails on:
Wanted but not invoked:
helloWorldService.handleInvocation(
"AccountA",
2023-08-14T08:22:29.100+01:00
);

When debugging you can see that the helloWorldService method in the test is invoked multiple times. Doing something similar in Micronaut Application with @MicronautTest will only trigger the helloWorldService method once, and the test succeeds

### Environment Information

MacOS Ventura 13.5
Micronaut version 4.0.3
Open JDK 17

### Example Application

_No response_

### Version

4.0.3

timyates:
This test fails prior to Micronaut 4.2.2 where we fixed it

See https://github.com/micronaut-projects/micronaut-core/pull/10264

Closes #1870


function-aws-test/src/test/java/io/micronaut/function/aws/test/EagerSingleton.java
```diff
@@ -4,4 +4,8 @@
 
 @Singleton
 public class EagerSingleton {
+
```
sdelamo, line 7: Can we use a different bean? I guess `EagerSingleton` is used in a different test. I would like to keep tests separate.

sdelamo: Moreover, should we have the been return something else like `return "goodby %s".formatted(name);` to verify a mock bean is actually used.

sdelamo (changes requested):

sdelamo (approved):

PR #2023, opened 2024-01-24
Remove http-netty dependency

sdelamo:


timyates:
Should we update servlet to 4.3.0 first?  It looks like there may be breaking changes in the servlet/core interface that need addressing

sdelamo:
> Should we update servlet to 4.3.0 first? It looks like there may be breaking changes in the servlet/core interface that need addressing

I think we can merge this PR

timyates (approved):

PR #2034, opened 2024-02-01
doc: improve paramterstore configuration reference

sdelamo:


timyates (approved):
I added some javadoc to setPrefixes too as that was missing in the configuration-properties description for that field.

I took the description from the super-class

Issue #2059, opened 2024-02-23
Problem JSON Infinite Recursion on Serialization


driverpt:
### Expected Behavior

To return proper Problem JSON

### Actual Behaviour

Goes into infinite loop

```
Caused by: io.micronaut.http.codec.CodecException: Error encoding object [https://zalando.github.io/problem/constraint-violation{400, Constraint Violation}] to JSON: Document nesting depth (1001) exceeds the maximum allowed (1000, from `StreamWriteConstraints.getMaxNestingDepth()`)
	at io.micronaut.json.codec.MapperMediaTypeCodec.encode(MapperMediaTypeCodec.java:233)
	at io.micronaut.servlet.http.ServletHttpHandler.encodeResponse(ServletHttpHandler.java:464)
	... 191 common frames omitted
Caused by: com.fasterxml.jackson.core.exc.StreamConstraintsException: Document nesting depth (1001) exceeds the maximum allowed (1000, from `StreamWriteConstraints.getMaxNestingDepth()`)
	at com.fasterxml.jackson.core.StreamWriteConstraints._constructException(StreamWriteConstraints.java:177)
	at com.fasterxml.jackson.core.StreamWriteConstraints.validateNestingDepth(StreamWriteConstraints.java:162)
	at com.fasterxml.jackson.core.json.UTF8JsonGenerator.writeStartObject(UTF8JsonGenerator.java:380)
	at io.micronaut.serde.jackson.JacksonEncoder.encodeObject(JacksonEncoder.java:105)
	at io.micronaut.serde.support.serializers.CustomizedObjectSerializer.serialize(CustomizedObjectSerializer.java:57)
	at io.micronaut.serde.support.serializers.CustomizedObjectSerializer.serializeIntoInternal(CustomizedObjectSerializer.java:157)
	at io.micronaut.serde.support.serializers.CustomizedObjectSerializer.serialize(CustomizedObjectSerializer.java:58)
	at io.micronaut.serde.support.serializers.CustomizedObjectSerializer.serializeIntoInternal(CustomizedObjectSerializer.java:157)
	at io.micronaut.serde.support.serializers.CustomizedObjectSerializer.serialize(CustomizedObjectSerializer.java:58)
	at io.micronaut.serde.support.serializers.CustomizedObjectSerializer.serializeIntoInternal(CustomizedObjectSerializer.java:157)
[...]

PR #2060, opened 2024-02-26
Fix serialization infinite recursion

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-aws/issues/2059


timyates (approved):
Approved, but I'm not sure why it was added 🤔 

@sdelamo do you agree this was probably a mistake?

sdelamo:
> Approved, but I'm not sure why it was added 🤔
> 
> @sdelamo do you agree this was probably a mistake?

Yes, it seems like a mistake.

driverpt:
What's the ETA of this Release?

PR #2131, opened 2024-05-31
Upgrade dependencies and fix build

sdelamo:


yawkat:
@sdelamo i pushed changes that, along with https://github.com/micronaut-projects/micronaut-servlet/pull/720 , should fix the failures introduced by the mn update

graemerocher (approved):

PR #2297, opened 2025-02-17
Update AWS to Core 4.8.x

sdelamo:


function-aws-api-proxy-test/build.gradle.kts
```diff
@@ -5,7 +5,6 @@ plugins {
 dependencies {
     api(mn.micronaut.http.server)
     api(projects.micronautFunctionAwsApiProxy)
-    implementation(platform(mnServlet.boms.jetty))
```
sdelamo, line 8: In this PR, I changed the build to no longer depend on the version of jetty from servlet. The `micronaut-function-aws-api-proxy-test` code assumes Jetty is still 11. We should update the code to Jetty 12. But I suggest doing it in a different PR.

sdelamo (commented):
There are some tck failures

PR #2315, opened 2025-03-19
Use Built-in Java HTTP Server instead of Jetty 11

sdelamo:
Jetty 11 has a CVE. This pull request removes the usage of Jetty 11 and replaces it with the built-in Java HTTP Server support introduced in servlet-core. 

Jetty 12 Handler uses `org.eclipse.jetty.server.Request` instead of  `jakarta.servlet.http.HttpServletRequest`. `org.eclipse.jetty.server.Request` no longer implements `HttpServletRequest`. Because of that, to get rid of Jetty 11, we had two options: rewrite to Jetty 12 or migrate `AwsApiProxyTestServer` to use the [Java Built-in HTTP Server](https://docs.oracle.com/javase/8/docs/jre/api/net/httpserver/spec/com/sun/net/httpserver/HttpServer.html) and remove the usage of Jetty. 

I gone the path of using the built-in HTTP Server. There are several classes introduced in this PR which I think should be in micronaut core.


function-aws-api-proxy-test/src/main/java/io/micronaut/function/aws/proxy/test/HttpServerEmbeddedServer.java
```diff
@@ -0,0 +1,118 @@
...
+import io.micronaut.http.server.exceptions.HttpServerException;
+import io.micronaut.http.server.exceptions.ServerStartupException;
+import io.micronaut.runtime.ApplicationConfiguration;
+import io.micronaut.runtime.server.EmbeddedServer;
+
+import java.net.*;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+@Internal
+class HttpServerEmbeddedServer implements EmbeddedServer {
```
sdelamo: @graemerocher @yawkat I would like to move this class to core. What do you think? That way I could use it in every cloud module (aws, gcp, azure).

sdelamo: or maybe in servlet.

yawkat: +1 to putting it into servlet

test-suite-http-server-tck-function-aws-api-proxy-test/src/test/java/io/micronaut/http/server/tck/lambda/tests/MicronautLambdaHandlerHttpServerTestSuite.java
```diff
@@ -11,9 +11,8 @@
     "io.micronaut.http.server.tck.lambda.tests"
 })
 @ExcludeClassNamePatterns({
-        "io.micronaut.http.server.tck.tests.forms.FormUrlEncodedBodyInRequestFilterTest",
-        "io.micronaut.http.server.tck.tests.forms.FormsSubmissionsWithListsTest",
-        "io.micronaut.http.server.tck.tests.filter.options.OptionsFilterTest",
+    "io.micronaut.http.server.tck.tests.OctetTest",
```
yawkat: why does OctetTest of all things fail?

sdelamo: it seems they payload fails base64 decoding: 

```
15:17:46.851 [HTTP-Dispatcher] ERROR i.m.f.a.p.test.AwsProxyHttpHandler - Illegal base64 character d
java.lang.IllegalArgumentException: Illegal base64 character d
```

https://github.com/micronaut-projects/micronaut-aws/blob/built-in-http-server/function-aws-api-proxy-test/src/main/java/io/micronaut/function/aws/proxy/test/AwsProxyHttpHandler.java#L70

sdelamo:
@yawkat @graemerocher can we merge this?

yawkat:
No, I want to figure out the octet thing first

yawkat (approved):

yawkat:
@sdelamo not sure why we use the mime encoder here, i couldnt find any docs requiring it https://github.com/micronaut-projects/micronaut-aws/blob/e18f645997a75a6093e7742bbf85cb09004dfe58/function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/payload2/APIGatewayV2HTTPResponseServletResponse.java#L53

PR #2317, opened 2025-03-21
feat(function-aws-api-proxy): add multipart/form-data support to ApiGatewayServletRequest

### LINKED ISSUE #270: Support for multipart upload for API Gateway
It would be nice to have support for file uploads with API Gateway out of the box. #269
adds basic support for handling binary formats with an example of how to handle file uploads manually using https://github.com/javadelight/delight-fileupload

nanodeath:
Hi there! This is a change to add support for multipart form uploads, e.g. file uploads. The `Serverless Function` mode already supports file uploads, but if you want to use them in your controllers (`Application` mode), we need something like this.

Previously, this module was using the query string decoder on the body for multipart uploads, which isn't right -- multipart form bodies have their own crazy syntax.

This adds support to the base ApiGatewayServletRequest for multipart/form-data requests and argument bindings for:
* `@Part String name` (binding to a part of name `name`)
* `@Part("foo") String name`
* `@Part("file") CompletedFileUpload file`

This does NOT add support for:
* Streaming file uploads (which I'm not sure we can easily support anyway)

This change is currently missing:
* Unit tests for the new classes
* Integration tests (?) for handling API Gateway events containing file uploads.
  * If this is something we need, I'd appreciate any pointers you have on how to implement this.

I've created a simple test project that can verify this works and can provide if there's interest.

Looking for feedback and direction on things that need to be implemented before this can be merged. Thanks!
* Feedback on thread safety
* Feedback on whether there's a better way to store file uploads -- they don't exactly fit in the parameters map, which is a String map.
* Feedback on class visibility, package structure, annotations

Closes #270.

### DIFF
--- a/function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/ApiGatewayBinderRegistry.java
+++ b/function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/ApiGatewayBinderRegistry.java
@@ -18,6 +18,9 @@ package io.micronaut.function.aws.proxy;
 import io.micronaut.context.annotation.Replaces;
 import io.micronaut.core.annotation.Internal;
 import io.micronaut.core.convert.ConversionService;
+import io.micronaut.function.aws.proxy.multipart.CompletedFileUploadBinder;
+import io.micronaut.function.aws.proxy.multipart.PartAnnotationRequestArgumentBinder;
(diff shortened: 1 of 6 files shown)


graemerocher:
Nice! Will review next week

Gustl22:
@graemerocher Ping for review. Thanks for your time.

PR #2454, opened 2025-12-18
Lambda API Gateway V2 Handler - Fix Payload Format

baylorpaul:
When using API Gateway V2, some information is not transmitted in the expected payload format.
See https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html#http-api-develop-integrations-lambda.proxy-format

This meant I was unable to transmit data in a few ways, including sending multiple cookies. This has been addressed in:
1. `CustomAPIGatewayV2HTTPResponseServletResponse.java`: Stop setting "multiValueHeaders". Set "headers" with comma separated values. Move "Set-Cookie" headers to "cookies".
2. `CustomAPIGatewayV2HTTPEventHandler.java`: Ensure the API Gateway V2 "cookies" are provided as a request header.
3. `CustomAPIGatewayV2HTTPEventServletRequest.java`: Support multiQueryParameters. E.g. `?parameter1=value1%2Cvalue2&parameter1=value3`.

I have provided a commented out test class (`ApiGatewayV2Test.java`) which I used to test these scenarios. Some of these tests verify that parameters reach the Micronaut controller with the same results and the response is identical whether making a direct API request or with an API Gateway V2 handler, which is HTTP abstraction (not a pass-through) that alters the data before and after reaching the Lambda. I have also verified these results with deployments to AWS.

Of additional note: I see that `ApplicationLoadBalancerServletRequest` and `ApiGatewayProxyServletRequest` have similar code. They may also require similar updates.


graemerocher:
The PR comments out all the tests, that probably needs to be fixed first

baylorpaul:
Sure - I've now uncommented out the test class and added the necessary dependencies for the tests to run

baylorpaul:
FYI - I have now additionally tested `io.micronaut.function.aws.proxy.payload2.APIGatewayV2HTTPEventFunction` with Lambda Function URLs. Lambda Function URLs have the same format as API Gateway V2:
- https://docs.aws.amazon.com/lambda/latest/dg/urls-invocation.html
- https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html#http-api-develop-integrations-lambda.proxy-format

Having said that, it may be worth adding some documentation that it is the appropriate handler for Lambda Function URLs as well. And possibly rename the handler to something more generic or making an identical handler for Lambda Function URLs.

baylorpaul:
Following up on this. Could we merge this in?

baylorpaul:
I have now rebased the branch on top of Micronaut v5. It appears that parameter binding has been altered between Micronaut v4 & v5, so I updated the test for that so that the API Gateway request (or Lambda Function URL request) still matches a direct request.

sdelamo:
@baylorpaul For payload v2 you should use: `RequestHandler<APIGatewayV2HTTPEvent,APIGatewayV2HTTPResponse>`. Is that not working?

> `micronaut-function-aws-api-proxy` modules supports both Payload format version v1 and v2. For functions of type Application, use the handler [ApiGatewayProxyRequestEventFunction](https://micronaut-projects.github.io/micronaut-aws/5.0.0/api/io/micronaut/function/aws/proxy/payload1/ApiGatewayProxyRequestEventFunction.html) for [Payload format version 1.0](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html#http-api-develop-integrations-lambda.proxy-format) and [APIGatewayV2HTTPEventFunction](https://micronaut-projects.github.io/micronaut-aws/5.0.0/api/io/micronaut/function/aws/proxy/payload2/APIGatewayV2HTTPEventFunction.html) for 2.0.

Do you have a sample app and instructions to reproduce the issue?

gradle/libs.versions.toml
```diff
@@ -6,6 +6,7 @@ bouncycastle = '1.70'
 fileupload = '0.0.8'
 logback-json-classic = '0.1.5'
 
+micronaut-data = "5.0.4"
```
sdelamo, line 9:
```suggestion
```

gradle/libs.versions.toml
```diff
@@ -43,6 +44,7 @@ micronaut-gradle-plugin = "5.0.1"
 javapoet = { module = "com.squareup:javapoet", version.ref = "javapoet" }
 
 micronaut-core = { module = 'io.micronaut:micronaut-core-bom', version.ref = 'micronaut' }
+micronaut-datajdbc = { module = "io.micronaut.data:micronaut-data-jdbc", version.ref = "micronaut-data" }
```
sdelamo, line 47:
```suggestion
```

sdelamo (changes requested):
I don't think the Micronaut data dependencies are necessary

sdelamo:
@baylorpaul I found a bug binding "Duplicate query strings" with `QueryValue` and `List<String>` https://github.com/micronaut-projects/micronaut-aws/pull/2544 Anything else you found broken?

sdelamo:
@baylorpaul is there anything not covered by https://github.com/micronaut-projects/micronaut-aws/pull/2544 ?

baylorpaul:
This PR is still necessary. #2544 affected the unit test results, but is ultimately a separate issue.

This PR deals with the AWS layer, `List` values from HTTP requests with commas (not arrays), and cookies with multiple values.

As you know with the AWS layer, raw HTTP requests are not passed directly to the Micronaut server. They are interpreted via the AWS payload & translated back for Micronaut. API Gateway V2 provides comma-separated values for "queryStringParameters", so we must use its "rawQueryString" to maintain consistency (see [Payload format version](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html#http-api-develop-integrations-lambda.proxy-format)). This is true for both "API Gateway V2 Handler" and the newer "Lambda Function URLs", which use the same format.

With `List` values especially, there is an interpretation issue when commas are introduced. Separately, there are also cookie issues when a cookie has multiple values.

These issues are demonstrated in `ApiGatewayV2Test.java`. This PR is different than #2544, but I believe #2544 is why I needed to update my ApiGatewayV2 tests in my latest commit. They can be demonstrated with an app as well, but it's much quicker to run the tests than redeploy an app for each alteration. The tests specify the expected results and run it through:
1) a standard Micronaut app without an AWS layer
2) both the AWS layer and subsequently the Micronaut app

See the `assertParamValuesMatchAndResponseContainsParams()` unit test. The results should be the same for both of the above cases.

Adding to the complexity (necessarily) is that some objects have specialized binders. One example is `Pageable`, which is why I've added the micronaut-data dependency for the tests. The "sort" in `Pageable` expects e.g. "created,desc". Contrary to normal binding, even when there is only one "sort" parameter, that should NOT be broken into two values for the `List<Sort>`.

I have an app, but I don't have a sample app immediately ready to go. If more than the unit tests are necessary, I can find some time to do that.

Regarding `RequestHandler<APIGatewayV2HTTPEvent,APIGatewayV2HTTPResponse>`:
Yes, I am using that (`APIGatewayV2HTTPEventFunction`). I misspoke in my original comment. In my local project I have Custom* classes to make these PR changes work in the meantime.

sdelamo:
@baylorpaul can you create a separate PR with a failing test with tests similar to the ones introduced in 
https://github.com/micronaut-projects/micronaut-aws/pull/2544 ?

baylorpaul:
I have refactored the test class. It should be a lot closer to the format of #2544 now. That makes it easier to read & easier to test each case. Let me know if you need just the tests in a separate PR. By reverting the 4 `APIGatewayV2HTTP*` classes in this PR, you can see the failing tests.

baylorpaul:
I have rebased the branch on 5.1.x. And I have created a separate PR with the failing test cases only --> https://github.com/micronaut-projects/micronaut-aws/pull/2570

PR #2555, opened 2026-07-12
Add Apache HttpClient 5 support for AWS SDK v2

### LINKED ISSUE #2554: AWS SDK v2 sync clients fail to start — no Apache HttpClient 5 support since SDK 2.46 made it the default
## Expected Behavior

A sync AWS client (e.g. `S3Client`) is created out of the box using the SDK's
default HTTP client, with connection/proxy settings configurable via
`aws.apache-client.*`.

## Actual Behaviour

`micronaut-aws-sdk-v2` only ships HTTP-client factories for the legacy Apache
HttpClient 4.x client (`software.amazon.awssdk:apache-client`, package
`software.amazon.awssdk.http.apache`), the Netty async client, and URLConnection.
There is no factory for the Apache HttpClient 5.x client
(`software.amazon.awssdk:apache5-client`, package
`software.amazon.awssdk.http.apache5`).

Per AWS, since AWS SDK for Java v2 2.46.0 the SDK sync clients use Apache 5 as
the default and the service modules pull `apache5-client` transitively instead
of `apache-client` 4.x:

- Announcement — Apache 5 is now the default HTTP Client:
  https://github.com/aws/aws-sdk-java-v2/issues/7007
- GA release blog — HTTP client built on Apache HttpClient 5.6:
  https://aws.amazon.com/blogs/developer/ga-release-of-the-aws-sdk-java-2-x-http-client-built-on-apache-httpclient-5-6/
- Configure the Apache 5.x based HTTP client (dev guide):
  https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/http-configuration-apache5.html

The Micronaut AWS 5.1.x BOM is already at 2.46.19:
https://github.com/micronaut-projects/micronaut-aws/blob/01614afa1/gradle/libs.versions.toml

Because the SDK clients are declared `compileOnly`, the effective runtime client
is whichever the consumer's service artifacts bring in:
https://github.com/micronaut-projects/micronaut-aws/blob/01614afa1/aws-sdk-v2/build.gradle.kts#L12-L14

On the SDK 2.46 default classpath (apache5-client present; apache-client 4.x and
url-connection-client absent):

- `ApacheClientFactory` / `ApacheClientConfiguration` beans do not load. The
  package is gated on `@Requires(classes = ApacheHttpClient.class)`:
  https://github.com/micronaut-projects/micronaut-aws/blob/01614afa1/aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/client/apache/package-info.java#L22
  and the config references `software.amazon.awssdk.http.apache.ApacheHttpClient`,
  which is absent:
  https://github.com/micronaut-projects/micronaut-aws/blob/01614afa1/aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/client/apache/ApacheClientConfiguration.java#L22
- `UrlConnectionClientFactory` is only a fallback —
  `@Requires(missingBeans = SdkHttpClient.class)`:
  https://github.com/micronaut-projects/micronaut-aws/blob/01614afa1/aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/client/urlConnection/UrlConnectionClientFactory.java#L47
  and its `UrlConnectionHttpClient` class is also absent by default.
- Result: no sync `software.amazon.awssdk.http.SdkHttpClient` bean exists.
  `AwsClientFactory#syncBuilder(SdkHttpClient)` requires one and calls
  `.httpClient(httpClient)` on the builder:
  https://github.com/micronaut-projects/micronaut-aws/blob/01614afa1/aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/service/AwsClientFactory.java#L82-L84
  so sync service client beans (`S3Client`, `SqsClient`, ...) cannot be created.

## Steps To Reproduce

1. Application with `io.micronaut.aws:micronaut-aws-sdk-v2` and
   `software.amazon.awssdk:s3`, no explicit HTTP-client dependency, AWS SDK >= 2.46.
2. Inject an `S3Client`.
3. Startup fails — no `SdkHttpClient` bean available for the sync builder.

## Environment Information

- Micronaut AWS 5.1.x, AWS SDK for Java v2 >= 2.46.0 (BOM 2.46.19)
- JDK 25

## Example Application

(link a minimal repro if requested)

## Version

5.1.0

MariusVolkhart:
## Summary

AWS SDK for Java v2 makes `apache5-client` (Apache HttpClient 5.x) the transitive sync
default as of `2.46.0`. This module previously wired only the legacy `apache-client` 4.x
factory, so an application on the default classpath had no sync `SdkHttpClient` bean and its
sync service clients (`S3Client`, `SqsClient`, …) failed to start.

Fixes #2554.

## What this adds

- A `io.micronaut.aws.sdk.v2.client.apache5` package — `Apache5ClientConfiguration` and
  `Apache5ClientFactory` — mirroring the existing `client.apache` package and producing the
  sync `SdkHttpClient` from `apache5-client`. The dependency is `compileOnly`, so the module
  adds no HTTP client transitively.
- Documentation for the 5.x client and updated URLConnection exclusion / `http.service.impl`
  guidance.

## Configuration is unchanged for users

The 4.x and 5.x clients bind the same `aws.apache-client.*` properties, so an application
needs no configuration change to move between them.

## One sync client per classpath

The AWS SDK rejects more than one sync HTTP client implementation on the classpath
(`Multiple HTTP implementations were found on the classpath`). Applications depend on exactly
one of `apache5-client` or `apache-client`. Because the two cannot coexist, they are verified
in separate test source sets: the main suite runs against `apache5-client`, and a
`testApache4` source set runs the legacy specs against `apache-client`.

## Compatibility

Non-breaking. The legacy `ApacheClientFactory` is unchanged; the 4.x client behaves exactly
as before when it is the sole sync client on the classpath.


sdelamo:
Thanks for the contribution @MariusVolkhart I merged your PR here: https://github.com/micronaut-projects/micronaut-aws/pull/2567

PR #2576, opened 2026-08-20
Guard AWS CRT beans on the optional async CRT client

n0tl3ss:
The AWS CRT factory is an optional integration, but its generated bean definitions reference both the synchronous and asynchronous CRT clients. Guarding only package configuration does not propagate the condition to those generated factory definitions, so native startup can resolve a missing optional async CRT type.

This moves the condition onto AwsCrtClientFactory, where Micronaut applies it to every generated factory bean definition. The regression test inspects the generated definition metadata and verifies both optional CRT classes are required; it fails when the factory annotation is removed.

Tests: ./gradlew :micronaut-aws-sdk-v2:test --no-daemon (50 tests).

### DIFF
--- a/aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/client/crt/AwsCrtClientFactory.java
+++ b/aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/client/crt/AwsCrtClientFactory.java
@@ -19,6 +19,7 @@ import io.micronaut.context.annotation.Bean;
 import io.micronaut.context.annotation.BootstrapContextCompatible;
 import io.micronaut.context.annotation.Factory;
 import io.micronaut.context.annotation.Prototype;
+import io.micronaut.context.annotation.Requires;
 import io.micronaut.core.annotation.Internal;
 import jakarta.inject.Singleton;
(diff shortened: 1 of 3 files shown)


graemerocher (approved):

n0tl3ss:
The failing Java CI tests showed `AwsCrtClientConfiguration` was still eagerly loaded when the optional CRT dependency was absent (`NoClassDefFoundError: AwsCrtHttpClient`).

Commit `1b86862c` keeps the narrow factory guard and restores the package-level `@Requires` for both optional CRT client classes. Added a regression test that starts an ApplicationContext without the CRT dependency; the focused AWS SDK v2 test now passes 2/2. The existing GraalVM native test suites remain unchanged and will rerun in CI.

sdelamo (approved):

n0tl3ss:
Crema/native rationale: native closed-world bean analysis can retain an optional AWS CRT reference even when the CRT implementation is absent, while JVM class loading resolves the optional type lazily. The conditional bean guard prevents native startup failure without changing the normal JVM path.

Observed native symptom: generated AWS CRT bean definitions retained references to the optional asynchronous CRT client, so native startup could resolve a missing optional CRT type. JVM class loading resolves that optional type lazily.

PR #2586, opened 2026-09-09
Guard Apache SDK client when HTTP implementation is selected

n0tl3ss:
The default Apache SDK client factory must not activate when another HTTP implementation is selected through the AWS SDK system property. Require the property to be absent so the property-specific factories remain mutually exclusive.

Tests:
`./gradlew :test-suite-groovy-aws-sdk-v2-http-client-apache4:test --tests io.micronaut.aws.sdk.v2.client.apache4.ApacheClientSystemPropertySpec --rerun-tasks --stacktrace`

### DIFF
--- a/aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/client/apache/ApacheClientFactory.java
+++ b/aws-sdk-v2/src/main/java/io/micronaut/aws/sdk/v2/client/apache/ApacheClientFactory.java
@@ -42,7 +42,7 @@ public class ApacheClientFactory {
      */
     @Bean(preDestroy = "close")
     @Singleton
-    @Requires(property = UrlConnectionClientFactory.HTTP_SERVICE_IMPL, notEquals = UrlConnectionClientFactory.URL_CONNECTION_SDK_HTTP_SERVICE)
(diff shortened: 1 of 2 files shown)


n0tl3ss:
Update: this is a framework-level overlap between two `@Requires` conditions,
not a native-only behavior. The current root-cause explanation and plain
`ApplicationContext` regression are documented in the follow-up comment below.

graemerocher:
This seems a bug in class path creation, why are there two candidates is what needs to be answered, not implementing workarounds to the fact there are two candidates

n0tl3ss:
Root cause: the two factory conditions overlap when
`software.amazon.awssdk.http.service.impl` explicitly selects Apache.

* `apacheClient()` currently uses `notEquals = URL_CONNECTION...`, which is
  true for the Apache service value.
* The property-specific Apache factory also matches that Apache service value.

Both methods therefore create an `SdkHttpClient`, producing the
`NonUniqueBeanException`. This is reproducible in a plain `ApplicationContext`;
it is not specific to native execution or classpath construction. The added
regression sets the Apache service property, fails before the change with two
candidates, and passes after the default factory is limited to a missing
property.

